PR #2, opened 2024-04-04
ON-15231: Proposal for solving the junit dependency issue

krishd-amd:
Successful run:

https://github.com/krishd-amd/tcpdirect/actions/runs/8725684997

For personal runs, please do ping me and I will share them.

With the above, don't really need the whole junit filtering given there are only 4 script tests that need to be checked.

### DIFF
--- a/.github/workflows/perform_build.yml
+++ b/.github/workflows/perform_build.yml
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: BSD-2-Clause
+# X-SPDX-Copyright-Text: (c) Copyright 2024 Advanced Micro Devices, Inc.
+
+name: "perform_build"
+
+on: push
+
+jobs:
+  extract_branch_info:
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    outputs:
+      tcpdirectBranch: ${{ steps.tcpdirectBranch.outputs.branch }}
+      onloadBranch: ${{ steps.onloadBranch.outputs.branch }}
+      packetDrillBranch: ${{ steps.packetDrillBranch.outputs.branch }}
+    steps:
+      - name: tcpdirect
+        uses: actions/checkout@v4
+        with:
+          path: tcpdirect
+
+      - name: Install yq
+        run: |
+           sudo apt update
+           sudo apt install python3 python3-pip jq
+           pip3 install yq
+
+      - name: Extract TCPDirect branch name
+        id: tcpdirectBranch
+        run: echo "branch=$(yq -r '.products.TCPDirect.version' < $GITHUB_WORKSPACE/tcpdirect/versions.yaml)" >> "$GITHUB_OUTPUT"
+
+
+      - name: Extract onload branch name
+        id: onloadBranch
+        run: echo "branch=$(yq -r '.products.Onload.version' < $GITHUB_WORKSPACE/tcpdirect/versions.yaml)" >> "$GITHUB_OUTPUT"
+
+      - name: Extract packetDrill branch name
+        id: packetDrillBranch
+        run: echo "branch=$(yq -r '.products.Packetdrill.version' < $GITHUB_WORKSPACE/tcpdirect/versions.yaml)" >> "$GITHUB_OUTPUT"
+
+
+  perform_build_and_test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    needs: extract_branch_info
+    steps:
+
+      - name: tcpdirect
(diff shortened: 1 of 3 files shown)


.github/workflows/perform_build.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Extract TCPDirect branch name
+        id: tcpdirectBranch
+        run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> "$GITHUB_OUTPUT"
+
+      - name: Extract Onload branch name
+        id: onloadBranch
+        run: |
+          export tcpdirect_branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
+          echo "branch=$(sed s/tcpdirect/onload/g <<< $tcpdirect_branch)" >> "$GITHUB_OUTPUT"
```
jfeather-amd, line 29: Might we want to build against the `last_known_good` branches that are known to at least build onload? That being said, they may be less useful if we're able to maintain branches in a good state by using actions to ensure things build and pass tests before merge.

.github/workflows/perform_build.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          repository:  ${{ github.repository_owner }}/packetdrill-tcpdirect
+          path: packetdrill-tcpdirect
+          ssh-key: ${{ secrets.PACKET_DRILL_PRIVATE_KEY }}
+          ref: tcpdirect-8.1
+
+      - name: tests
+        run: |
+          cd $GITHUB_WORKSPACE/tcpdirect
+          ONLOAD_TREE=$GITHUB_WORKSPACE/onload_internal NDEBUG=1 ZF_DEVEL=1 TEST_THREAD_NAME=zf make -j $(nproc) test
```
jfeather-amd, line 75: I think I would probably like to have `-k` here, so we run through all tests even if one fails. Although I expect this might make two aspects a bit more tricky:
1. Would the run still "fail" if we ignore test failures and continue on to run everything?
2. How discoverable are the test failures - is there a summary at the end, or something that can be (relatively) easily searched for (e.g., "not ok")?

jfeather-amd (commented):
Overall happy with these changes, I've added some comments but understand this is a draft so perhaps these are things you already had in mind. I think the only other comment I have is on this proposal, which I also mentioned in the meeting earlier today, is that I would probably want onload changes to trigger (at least) a tcpdirect build. Otherwise I'd worry that a totally unrelated PR (or otherwise) might see the fallout of a build breakage from the onload side.

.github/workflows/perform_build.yml
tcrawley-xilinx, line 29: I feel like this should be using `versions.yaml`

tcrawley-xilinx (commented):
I'd like to see an example of an actual test failure (both on a github action and a jenkins pipeline)

.github/workflows/perform_build.yml
krishd-amd: This may be best as a future change for now, would rather this go in then sit here.

src/tests/zf_unit/Makefile.inc
```diff
@@ -8,15 +8,16 @@
 #    allowing to run subsets of tests with various variants
 UT_LOOP := zftest zfinit zfudprx zf_superbuf_reuse zftcprx zftcpcopyrx \
            zftcptx zfmulticast zftcp_muxer \
-           zfds efrxtimestamping zftimestamping zftcprx_on14495
+           zfds efrxtimestamping zftcprx_on14495
 UT_LOOP2 := zftest zfinit zfudprx zf_superbuf_reuse zftcprx zftcpcopyrx \
             zftcptx zfmulticast zftcp_muxer \
             zfds efrxtimestamping zftcprx_on14495
+UT_JENKINS := zfstackdump zftimestamping
```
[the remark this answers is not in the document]
krishd-amd: I think I have discussed this in person. Happy to adjust things later but for now this is necessary to have a green pipeline.

.github/workflows/perform_build.yml
jfeather-amd: Fair enough, definitely would like this in the future but agree that this can be iterated on.

jfeather-amd (approved):
Happy to have this merged with the latest changes. Also checked Krish's fork for the [latest run of these actions](https://github.com/krishd-amd/tcpdirect/actions/runs/8725684997/job/23939237817).

PR #3, opened 2024-04-04
ON-15668: add time checking for MAC lookup to better enforce ARP timeout

rdrinkwa-xilinx:
Existing logic assumed zero-cost route resolution, which was incorrect. New method uses clock time to identify when the ARP timeout has been exceeded.

### DIFF
--- a/src/lib/zf/cplane.c
+++ b/src/lib/zf/cplane.c
@@ -182,7 +182,9 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
   struct cp_fwd_data data;
   cicp_verinfo_t verinfo;
   int rc;
-  int iter;
+  int64_t time_limit_us;
+  int64_t time_used_us = 0;
+  struct timespec start;
   bool is_multi = CI_IP_IS_MULTICAST(path->dst);
 
   if( is_multi )
@@ -206,8 +208,14 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
                       CI_IP_PRINTF_ARGS(&key.dst.ip4),
                       key.ifindex);
 
-  iter = wait ? sti->arp_reply_timeout : 0;
+  time_limit_us = wait ? sti->arp_reply_timeout : 0;
+
+  if( time_limit_us )
+    clock_gettime(CLOCK_MONOTONIC, &start);
+
   do {
+    struct timespec now;
+
     rc = oo_cp_route_resolve(&zf_state.cplane_handle,
                              &verinfo, &key, &data);
     if( rc < 0 || data.base.ifindex != sti->sti_ifindex ) {
@@ -231,10 +239,16 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
     /* In theory, we shoud break out if we see FLAG_ARP_FAILED popping up
      * in a non-first loop.  But if we see ARP_FAILED from the start, we
      * should give a chance to re-resolve.  For now we always loop. */
-    if( (data.flags & CICP_FWD_DATA_FLAG_ARP_VALID) || iter == 0 )
+    if( (data.flags & CICP_FWD_DATA_FLAG_ARP_VALID) || (time_limit_us == 0) )
       break;
 
-    iter--;
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    time_used_us = ((now.tv_sec - start.tv_sec) * 1000000) +
+                   ((now.tv_nsec - start.tv_nsec) / 1000);
+
+    if( time_used_us >= time_limit_us )
+      break;
+
     usleep(1);
   } while( 1 );
 
@@ -243,7 +257,7 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
       zf_log_cplane_err(st, "ARP failed for " CI_IP_PRINTF_FORMAT "\n",
                         CI_IP_PRINTF_ARGS(&path->dst));
     }
-    else if( wait ) {
+    else if( time_limit_us ) {
       zf_log_cplane_err(st, "ARP timeout for " CI_IP_PRINTF_FORMAT "\n",
                         CI_IP_PRINTF_ARGS(&path->dst));
     }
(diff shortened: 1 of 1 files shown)


src/lib/zf/cplane.c
```diff
@@ -258,10 +274,10 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
                       CI_MAC_PRINTF_ARGS(&data.dst_mac),
                       data.base.ifindex, data.encap.type);
 
-  if( iter > 0 ) {
+  if( wait ) {
     zf_log_cplane_info(st, "ARP for " CI_IP_PRINTF_FORMAT
                        " via ifindex %d took %dus\n",
-                       CI_IP_PRINTF_ARGS(&path->dst), data.base.ifindex, iter);
+                       CI_IP_PRINTF_ARGS(&path->dst), data.base.ifindex, time_used);
```
pemberso-xilinx, line 280: Could you add some testing examples with this output just to show it all working with different attribute values.

src/lib/zf/cplane.c
```diff
@@ -182,7 +182,10 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
   struct cp_fwd_data data;
   cicp_verinfo_t verinfo;
   int rc;
-  int iter;
+  int time_limit;
+  int time_used;
```
abower-amd, line 186: Admittedly half an hour would be a long timeout but the limit of these integral types is a little close for comfort to times that could be seen in the real world, e.g. if you were using a simulation platform but measuring real time. I suggest a 64-bit quantity would be more appropriate for microseconds? Maybe use `time_t`?

src/lib/zf/cplane.c
```diff
@@ -231,10 +241,16 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
...
+
+    gettimeofday(&now, NULL);
+    time_used = ((now.tv_sec - start.tv_sec) * 1000000) +
+                (now.tv_usec - start.tv_usec);
+
+    if (time_used >= time_limit)
       break;
 
-    iter--;
     usleep(1);
```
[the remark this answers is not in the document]
rdrinkwa-xilinx: This would make the loop potentially very aggressive in polling the CP server, which may cause issues with other processes attempting to do the same. This looks to be limited to memory accesses though, since there are no blocking calls or locks I can see.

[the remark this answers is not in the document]
rdrinkwa-xilinx: I've opted to keep the sleep. This adds 1us on each iteration, which takes 22us so the additional time is small compared to a potentially aggressive polling.

rdrinkwa-xilinx:
Time for `zft_connect`  call to return if there's no route in microseconds:
```
timeout | old      | new
--------+----------+-------
-100    | n/a      | 43
0       | 44       | 22
1       | 114      | 42
10      | 606      | 43
20      | 1181     | 43
50      | 2797     | 72
100     | 5582     | 127
200     | 11150    | 222
500     | 27796    | 529
1000    | 55612    | 1021
2000    | 111132   | 2020
5000    | 277772   | 5019
10000   | 555537   | 10020
20000   | 1111074  | 20018
50000   | 2777214  | 50015
100000  | 5555256  | 100011
200000  | 11110579 | 199999
500000  | 27775830 | 499970
1000000 | 55551457 | 999916
```
-100 not available for the old code because it would spin for a long time in the loop due to incorrect interpretation.

src/lib/zf/cplane.c
```diff
@@ -182,7 +182,9 @@ __zf_cplane_get_path(struct zf_stack* st, struct zf_path* path,
   struct cp_fwd_data data;
   cicp_verinfo_t verinfo;
   int rc;
-  int iter;
+  time_t time_limit_us;
+  time_t time_used_us = 0;
```
jfeather-amd, line 186: I realise @abower-amd asked for `time_t` to be used, but is it safe to cast a signed integer to `time_t` given it isn't guaranteed to be signed (although is in seemingly most cases)? I expect this would only be very problematic for an unsigned 64-bit `time_t` (if one even exists), as a 32-bit unsigned `time_t` would only wait for a little over 4 seconds.

jfeather-amd (approved):
Looks like a reasonable change! I've left a comment, though it is more of a question about `time_t` than anything. Happy for this to be merged if you think the question is a non-issue.

src/lib/zf/cplane.c
mseymour-xilinx: Yes, there could be a problem if it's an unsigned type (unlikely) and `gettimeofday` jumps backwards (theoretically possible) giving a negative value. Better to use an explicitly signed type (long or int64_t) and/or use `clock_gettime(CLOCK_MONOTONIC)` to avoid negative times.

abower-amd: `time_t` is defined to be signed but yes the key thing is for it to be a 64-bit signed quantity however that is achieved and I agree CLOCK_MONOTONIC is the right way to do this.

mseymour-xilinx: `time_t` is implementation-defined, but we're quite strongly tied to a particular implementation so I suppose it doesn't really matter if we assume the definition is suitable for this useage. I'd still use an explicitly signed and sized type if that's what I wanted.

jfeather-amd: I think my concern was more with the possibility of a negative value being provided for `sti->arp_reply_timeout` which, when cast to `time_t`, could result in a massive `time_limit_us`, potentially many years long in the worst case of an unsigned implementation.

abower-amd: Sorry Mike, yes you are right the signedness isn't guaranteed. I agree this should simply be `int64_t` and `CLOCK_MONOTONIC` should be used.

rdrinkwa-xilinx:
Updated `time_t` to `int64_t` and replaced `gettimeofday()` with `clock_gettime(CLOCK_MONOTONIC)`.

mseymour-xilinx (approved):

PR #6, opened 2024-04-17
ON-15421: Don't export EF_VI symbols in the dynamic library

tcrawley-xilinx:
The trade sim had to be updated as it explicitly used ef_vi functions.
## Testing done
Built tcpdirect, ef_vi symbols no longer appear in the dynamic library.
```
$ nm -gD libonload_zf.so | grep ef
000000000028b018 D ci_log_prefix
000000000005c480 T ci_set_log_prefix
0000000000022efd T zf_stack_dump_summary
```


jfeather-amd (approved):

matthewr-xilinx:
Is this PR against the correct branch? Do we want to include this change in the next 8.1 update as well as on master?

PR #8, opened 2024-04-23
allowing setting log_level and log_format

andresa-xilinx:
log_level and log_format cannot be set by zf_attr_set(). Customer requested we return an error rather than allowing to seemingly succeed and not work.

However, by inspecting the zf_attr_set_int code and the unit test zfattrs.c, log_level and log_format never worked as intended. The unit test never showed the expected headers and logs when the attributes got set in the zfattrs.c function set_test_attrs.   

We do not see reasons to forbid programatic initialization of those parameters using zf_attr_set_int. The already existent unit test (zfattrs.c) works as expected after applying this patch.


andresa-xilinx:
Before applying the patch:
andresa@sup-cap1:~/onload-patches/tcpdirect$  ZF_ATTR="emu=2;interface=enp129s0f0" ./build/gnu_x86_64-zf-devel/bin/zf_unit/zfattrs
1..100
# attr test bitmask 0x20cca
ok 1 - Data sent
ok 2 - Correct data received
# attr test bitmask 0x5ab26
ok 3 - Data sent
ok 4 - Correct data received
# attr test bitmask 0x323f4
ok 5 - Data sent
ok 6 - Correct data received
# attr test bitmask 0x38dbd
ok 7 - Data sent
ok 8 - Correct data received
# attr test bitmask 0x6927f
ok 9 - Data sent
ok 10 - Correct data received
# attr test bitmask 0x1a22a
ok 11 - Data sent
ok 12 - Correct data received
# attr test bitmask 0x2cfd4
ok 13 - Data sent
ok 14 - Correct data received
# attr test bitmask 0x4b20
ok 15 - Data sent
ok 16 - Correct data received
# attr test bitmask 0x50269
ok 17 - Data sent
ok 18 - Correct data received
# attr test bitmask 0x3348e
ok 19 - Data sent
ok 20 - Correct data received
# attr test bitmask 0x71b2f
ok 21 - Data sent
ok 22 - Correct data received
# attr test bitmask 0x1bf39
ok 23 - Data sent
ok 24 - Correct data received
# attr test bitmask 0x293d3
ok 25 - Data sent
ok 26 - Correct data received
...
# attr test bitmask 0x4627a
ok 95 - Data sent
ok 96 - Correct data received
# attr test bitmask 0xc369
ok 97 - Data sent
ok 98 - Correct data received
# attr test bitmask 0x5620d
ok 99 - Data sent
ok 100 - Correct data received

andresa-xilinx:
After applying the patch:
```
andresa@sup-cap1:~/onload-patches/tcpdirect$  ZF_ATTR="emu=2;interface=enp129s0f0" ./build/gnu_x86_64-zf-devel/bin/zf_unit/zfattrs
1..100
# attr test bitmask 0x7c5ae
ok 1 - Data sent
ok 2 - Correct data received
# attr test bitmask 0x247ed
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 3 - Data sent
ok 4 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x6c2d
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 5 - Data sent
ok 6 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x26443
[zfattrs:805298]      7555961057275990   1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 7 - Data sent
ok 8 - Correct data received
[zfattrs:805298]      7555961057778507 enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x15df7
[zfattrs:805298]      7555961087841461   1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 9 - Data sent
ok 10 - Correct data received
[zfattrs:805298]      7555961088777410 enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x46f65
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 11 - Data sent
ok 12 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x5f4a5
  1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 13 - Data sent
ok 14 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x1e569
  1000 | rxq_size=511 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 15 - Data sent
ok 16 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x71bc0
ok 17 - Data sent
ok 18 - Correct data received
# attr test bitmask 0x25398
ok 19 - Data sent
ok 20 - Correct data received
# attr test bitmask 0xdf8e
ok 21 - Data sent
ok 22 - Correct data received
# attr test bitmask 0xb43
[zfattrs:805298]      7555961304932210   1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 23 - Data sent
ok 24 - Correct data received
[zfattrs:805298]      7555961305442439 enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x77e0f
[zfattrs:805298]      7555961336022835   1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 25 - Data sent
ok 26 - Correct data received
[zfattrs:805298]      7555961336963651 enp129s0f0/000  1000 | zf_stack_free()
...
ok 90 - Correct data received
# attr test bitmask 0x622ad
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 91 - Data sent
ok 92 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x53295
  1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 93 - Data sent
ok 94 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x7259b
[zfattrs:805298]      7555962412005493   1000 | rxq_size=511 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 95 - Data sent
ok 96 - Correct data received
[zfattrs:805298]      7555962412724578 enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x44730
ok 97 - Data sent
ok 98 - Correct data received
# attr test bitmask 0x8da4
ok 99 - Data sent
ok 100 - Correct data received

```

pemberso-xilinx (commented):
Looks like a useful improvement.  Could you update this line in the doc too  https://github.com/Xilinx-CNS/tcpdirect/blob/834fa202e1f5027432cfb1c3f5023b6b44333849/src/lib/zf/doxygen/040_using.dox#L116.  Suggest saying to the effect that logging can be controlled via API after library initialisation but any logging from zf_init() will be controlled by env vars.

jfeather-amd (commented):
Looks like a nice change! I ran the unit tests locally and they still seem to be happy and ignore the additional logging output.

I think it would be nice if the unit tests actually verified that this specific option resulted in some extra output, perhaps with some wrapper script to parse the additional output while keeping the existing tests. Although I don't expect you to do this in this PR unless you particularly want to.

I do agree with @pemberso-xilinx though, that the documentation should be updated to reflect this change.

andresa-xilinx:
With this commit log_file can be set programatically. The unit test shows log_file gets set correctly and it reports the desired output (either in stderr or in a file). Note that as zf_log_file is static, when the attrs get freed within the app the current behavior is not to change log_file value.

I would have expected that when attrs gets freed, log_file should return back to the original value (stderr). This is to prevent new attrs set not to accidentally inherit the previous value in log_file. Therefore, I have also included this change in behavior.

andresa-xilinx:
Previous behavior was to log everything either to stderr or to a file. 
Now the app can choose among the two behaviors programatically as follows. See log messages in the console correspond to default behavior logging to stderr. Later we can see the listing of the unique directory and the different files created by different independent test iterations. 
 
```
andresa@sup-cap2:~/onload-patches/tcpdirect$ ZF_ATTR="emu=2;interface=enp129s0f0" ./build/gnu_x86_64-zf-devel/bin/zf_unit/zfattrs
1..100
# attr test bitmask 0xb5f28
ok 1 - Data sent
ok 2 - Correct data received
# attr test bitmask 0x4ebb2
ok 3 - Data sent
ok 4 - Correct data received
# attr test bitmask 0x1c3f4
ok 5 - Data sent
ok 6 - Correct data received
# attr test bitmask 0xe1a75
  1000 | rxq_size=511 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 7 - Data sent
ok 8 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0xe6051
  1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
ok 9 - Data sent
ok 10 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0xe1bcc
ok 11 - Data sent
ok 12 - Correct data received
# attr test bitmask 0x7e6fe
ok 13 - Data sent
ok 14 - Correct data received
# attr test bitmask 0xb34c4
ok 15 - Data sent
ok 16 - Correct data received
# attr test bitmask 0xcddf1
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 17 - Data sent
ok 18 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0x74e4c
ok 19 - Data sent
ok 20 - Correct data received
# attr test bitmask 0xbc299
...
# attr test bitmask 0x8d1c3
[zfattrs:624233]     10736589617248049   1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 91 - Data sent
ok 92 - Correct data received
[zfattrs:624233]     10736589617791237 enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0xa7b82
ok 93 - Data sent
ok 94 - Correct data received
# attr test bitmask 0x4adcf
ok 95 - Data sent
ok 96 - Correct data received
# attr test bitmask 0x415c5
  1000 | rxq_size=511 txq_size=511 evq_size=1014 rx_prefix_len=0
ok 97 - Data sent
ok 98 - Correct data received
enp129s0f0/000  1000 | zf_stack_free()
# attr test bitmask 0xdc1bc
ok 99 - Data sent
ok 100 - Correct data received
```

```
andresa@sup-cap2:~/onload-patches/tcpdirect$ ls /tmp/logfileBlqunM/ -l
total 44
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_0.log
-rw-r--r-- 1 andresa sflr 104 May  8 10:55 file_10.log
-rw-r--r-- 1 andresa sflr 105 May  8 10:55 file_11.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_12.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_13.log
-rw-r--r-- 1 andresa sflr 104 May  8 10:55 file_14.log
-rw-r--r-- 1 andresa sflr 183 May  8 10:55 file_15.log
-rw-r--r-- 1 andresa sflr 104 May  8 10:55 file_16.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_17.log
-rw-r--r-- 1 andresa sflr 104 May  8 10:55 file_18.log
-rw-r--r-- 1 andresa sflr 104 May  8 10:55 file_19.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_1.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_20.log
-rw-r--r-- 1 andresa sflr 183 May  8 10:55 file_21.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_22.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_2.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_3.log
-rw-r--r-- 1 andresa sflr 105 May  8 10:55 file_4.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_5.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_6.log
-rw-r--r-- 1 andresa sflr 105 May  8 10:55 file_7.log
-rw-r--r-- 1 andresa sflr 184 May  8 10:55 file_8.log
-rw-r--r-- 1 andresa sflr   0 May  8 10:55 file_9.log
andresa@sup-cap2:~/onload-patches/tcpdirect$ cat /tmp/logfileBlqunM/file_8.log
[zfattrs:624233]     10736588982406600   1000 | rxq_size=1023 txq_size=1023 evq_size=2038 rx_prefix_len=0
[zfattrs:624233]     10736588983238761 enp129s0f0/000  1000 | zf_stack_free()

```

src/tests/zf_unit/zfattrs.c
```diff
@@ -16,6 +16,7 @@ struct attr_value test_attrs[] = {
   { "log_level", 0x7777 },
   { "log_format", 15 },
   { "log_to_kmsg", 1 },
+  { "log_file", 0 },
```
pemberso-xilinx, line 19: For everything else, these are attribute values used in the test.  I think this could just be set to a single filename rather than using multiple files.  I'm not sure what additional value there is in testing with different files on different iterations rather than same file since there are no output checks here.

I've put further review comments below independent of above comment

src/tests/zf_unit/zfattrs.c
```diff
@@ -161,6 +172,12 @@ static void test(struct zf_stack* stack, struct zf_attr* attr,
 
 int main(void)
 {
+  char log_template[]="/tmp/logfileXXXXXX";
```
pemberso-xilinx, line 175: I don't see log_template reused so could just pass "/tmp/..." into mkdtemp

src/tests/zf_unit/zfattrs.c
```diff
@@ -34,12 +35,15 @@ struct attr_value test_attrs[] = {
   { "alt_count", 1 },
 };
 
+char *tmp_log_dir;
+int file_uid=0;
```
pemberso-xilinx, line 39: file_uid could just be static inside set_test_attrs rather than global

src/tests/zf_unit/zfattrs.c
```diff
@@ -55,7 +59,14 @@ static int set_test_attrs(struct zf_attr* attr)
     if( (random & (1 << i)) == 0 )
       continue;
 
-    rc = zf_attr_set_int(attr, test_attrs[i].name, test_attrs[i].value);
+    if( strcmp("log_file", test_attrs[i].name)==0 ) {
+      strcat(tmp_file, tmp_log_dir);
+      sprintf(&tmp_file[strlen(tmp_file)],"/file_%i.log",file_uid++);
```
pemberso-xilinx, line 64: would prefer strncat and snprintf for safety

pemberso-xilinx (changes requested):

src/lib/zf/attr.c
```diff
@@ -131,6 +131,7 @@ int zf_attr_alloc(struct zf_attr** attr_out)
     return -ENOMEM;
   if( zf_attr_set_from_sysctl(attr) )
     return -ENOSYS;
+  zf_log_stderr(); /*log_file is static and we want default behaviour for new set of attributes*/
```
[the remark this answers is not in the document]
andresa-xilinx: I'm mimicking the behavior in 
https://github.com/andresa-xilinx/tcpdirect/blob/345a7b452994c4cbe1cfbba27345f33abb124c18/src/lib/zf/zf.c#L20
Not sure whether we want to change this. I think log_file has a unique initialization procedure as shown in link above.

src/tests/zf_unit/zfattrs.c
andresa-xilinx: Sounds reasonable, but wanted to test the case of alternating between stderr and different files as per zf_attrs.c test. Moreover, it's what's done in the test, repeated tests randomly picking different attribute settings one of which is log_file which I believe it deserves to be independent on each iteration.

src/tests/zf_unit/zfattrs.c
andresa-xilinx: mkdtemp requires a char * and ISO C++ forbids converting const to char *

```
src/tests/zf_unit//zfattrs.c: In function ‘int main()’:
src/tests/zf_unit//zfattrs.c:175:50: error: ISO C++ forbids converting a string constant to ‘char*’ [-Werror=write-strings]
   tmp_log_dir = mkdtemp("/tmp/logfileXXXXXX");
                                                  ^
At global scope:
cc1plus: error: unrecognized command line option ‘-Wno-address-of-packed-member’ [-Werror]
cc1plus: all warnings being treated as errors
make: *** [Makefile:51: build/gnu_x86_64-zf-devel/obj/tests/zf_unit//zfattrs.o] Error 1

```

src/lib/zf/attr.c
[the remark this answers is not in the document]
andresa-xilinx: @pemberso-xilinx More precisely, the problem is that as log_file is global static, we need to make sure we close the previous file (in case it exists) before resetting the value to stderr. That is, ZF_ATTR macro in __zf_attr_reset() will only allow assigning the intended (stderr) value and would not be enough.

src/tests/zf_unit/zfattrs.c
pemberso-xilinx: ok.  I don't think it changes the effectiveness of the test either way.

src/tests/zf_unit/zfattrs.c
pemberso-xilinx: sorry, should have read the man page for mkdtemp. Didn't realise it modified the input.

pemberso-xilinx (changes requested):

src/lib/zf/attr.c
```diff
@@ -286,6 +287,14 @@ int zf_attr_set_str(struct zf_attr* attr, const char* name, const char* val)
       *p = strdup(val);
       if( *p == NULL )
         return -ENOMEM;
+      if( strcmp(name, "log_file")==0 ) {
+        int rc = zf_log_redirect(attr->log_file);
+        if( rc < 0 ) {
+          zf_log_stack_err(NO_STACK, "%s: Failed to redirect logging: %s\n",
+                          __func__, strerror(-rc));
+          /* Failure here is non-fatal. */
```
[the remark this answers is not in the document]
andresa-xilinx: Interesting, it seems a safer approach. Are you suggesting this should also be the case in zf_init()? that is terminating the initialization with an error, EPERM for example. 

https://github.com/andresa-xilinx/tcpdirect/blob/173a39aec69fc0d83c87af3837478700c3a4c64a/src/lib/zf/zf.c#L32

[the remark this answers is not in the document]
pemberso-xilinx: The logging output from `zf_init()` comes from environment rather than a direct API call to configure.  You could make a case for giving an error from `zf_init()` due to bad logging related env vars but I wouldn't change current behaviour now.

I do think this case where an API call with the single purpose of reconfiguring logging should return an error on failure though.  I wouldn't make it EPERM though as there are many reasons (ref `man 2 open`) for not being able to open file.  I would suggest more generic EINVAL. I advise getting a second opinion too.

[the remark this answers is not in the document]
abower-amd: > there are many reasons (ref man 2 open) for not being able to open file.

Wouldn't that suggest you should pass through the errno? Squashing errors to EINVAL is sometimes necessary but never nice!

abower-amd:
I think the next step is a re-review by @pemberso-xilinx to see if he is happy with the changes, thanks!

src/lib/zf/attr.c
```diff
@@ -126,6 +126,7 @@ static struct zf_attr* __zf_attr_alloc(void)
 int zf_attr_alloc(struct zf_attr** attr_out)
 {
   struct zf_attr* attr = __zf_attr_alloc();
+  zf_log_stderr(); /*log_file is static and we want default behaviour for new set of attributes*/
```
pemberso-xilinx, line 129: If an app creates different sets of attributes for different stacks, upon creating a second set of attributes it is going to change logging for the first since zf_log_file is global.

Not sure this is the behaviour we want.

Same applies to zf_log_level and zf_log_format.

Also note that zf_attr_dup() does not pass through here so it gives a path to create a new set of attributes with different behaviour to creating two sets manually.

I'm thinking a cleaner approach might be to replace global zf_log_level, etc with a global set of attributes that are only initialised in zf_init() and a call to modify any global attributes modifies only the global ones.  I don't think changing "local" attributes to then modify global values makes sense.  All the local ones could be different values with only one sets of attributes matching the global ones.

Sorry for not picking this up earlier but only thought through it when seeing this call here now.

src/lib/zf/attr.c
```diff
@@ -281,14 +282,25 @@ int zf_attr_set_str(struct zf_attr* attr, const char* name, const char* val)
...
+      if( strcmp(name, "log_file")==0 ) {
+        int rc = zf_log_redirect(val);
+        if( rc < 0 ) {
+          zf_log_stack_err(NO_STACK, "%s: Failed to redirect logging: %s\n",
+                          __func__, strerror(-rc));
+          /* Failure here is non-fatal. */
+          return rc;
+        }
+      }
+      free(*p);
```
pemberso-xilinx, line 295: both if/else paths call free(*p); so think this can be moved straight after get_field_str as before.

src/lib/zf/attr.c
```diff
@@ -281,14 +282,25 @@ int zf_attr_set_str(struct zf_attr* attr, const char* name, const char* val)
...
       *p = strdup(val);
       if( *p == NULL )
```
pemberso-xilinx, line 297: there is a small chance the strdup fails after changing the value of the attr holding the file name is out of sync with where it is actually going.  you would need to strdup to a temp val first and then assign *p to it later

pemberso-xilinx (changes requested):

src/lib/zf/attr.c
andresa-xilinx: zf_log_file, zf_log_level and zf_log_format were initially intended to be global before ON-15400 and the current behaviour is just inherited. If this is not the case, I think we should raise a separate case for this, or are you proposing to forbid the zf_attr_set_str( ) for log_file, and to forbid zf_attr_set_int( ) for log_level and log_format to be directly called from the user's code?
I'm a bit confused as the proposed code in this PR (ON-15400) do not change "local" attributes, it changes the global attributes as originally intended, but basically the code doing that change was just missing and, now, this PR adds it.

pemberso-xilinx: Currently (prior to any of your changes) all calls to zf_attr_set_...(attr) operate only on attributes within attr and different stacks can use different attributes.  The log_file, log_format, log_level are global attributes and changing them will modify these settings across all stacks not only for a stack related to the attr it is called on.  I think this is ok if well explained / documented.

What I do think we need to change is that when zf_attr_alloc() is called, logging changes will get reset and I think this would be unexpected.  I.e. this sequence

1. zf_attr_alloc(&attr1)
2. zf_attr_set_...(attr1, <change logging settings>)
3. zf_stack_alloc(attr1, &stack1)
4. zf_attr_alloc(&attr2) <=== at this point I expect the global logging options already configured to get reset with the current implementation.  Instead I think global attributes should only be initialised when calling zf_init() or when explicitly modified with a zf_attr_set_...() call.

At the least we should have tests for above sequence to check the behaviour and whether my interpretation is correct.

Also consider the sequence when in step 4. you create attr2 by calling zf_attr_dup() on attr1.  This follows a different code path and would currently give a different behaviour to two calls to zf_attr_alloc() with regards to resetting logging options which I don't think should be the case.

andresa-xilinx: Yes, agreed. Missing a bit of documentation for the new behaviour when modifying log_file, log_format and log_level as these are global attributes.
Also, it is true that allocating a new set of attributes (via zf_attr_alloc) will reset the global log_file to stderr and the proposed sequence of calls end up in an unintended behaviour. The problem is that the program line 129 referenced above should not be there. Removing it fixes the problem.

Also, checked zf_attr_dup and it does respect the new behaviour, ie, although it does initialize the log_file to stderr at allocation time, and copy the log_file value from the source "attr", it does not modifies the zf_log_file  static value. Hence, the existent value in zf_log_file remains intact after calling zf_attr_dup. Good test to consider, though. 

```
struct zf_attr* zf_attr_dup(const struct zf_attr* from)
{
  struct zf_attr* to = __zf_attr_alloc();
  if( to != NULL )
    zf_attr_copy(to, from);
  return to;
}  
```

src/lib/zf/attr.c
andresa-xilinx: done. Moved to where it was before.

src/lib/zf/attr.c
andresa-xilinx: I opted for a slightly different solution where code check first the potential failure of zf_log_redirect and then the potential failure of strdup. In this way, it saves the app from performing the most expensive operation first (strdup) and having to eventually free the resources again if the zf_log_redirect were to fail afterwards.

ligallag-amd (commented):
Given that `zf_log_level`, `zf_log_format` and `zf_log_file_name` are all globals, I don't see why we don't have getters and setters for them rather than this unusual juggling where the zf_attr isn't actually mutated and global state is.

I also wonder if this current interface may lead the programmer to think this interface is thread safe which I don't think it is by the nature of shared global data. This begs the question of how these globals should be synchronised. You could add a lock which can be acquired which signals that these globals may be modified.

andresa-xilinx:
@ligallag-amd Yes, I also wondered the same about having getters and setters. But see, there already is a single (pre existent) getter: 

`static uint64_t zf_log_level_get()
`
Which is not documented as such in the TCP Direct user's guide, nor do I think was the intention of that API.   However, it is exactly that, a standard getter only used internally within TCP Direct. 

I wonder what the implications would be with customers if we now change the API and forbid the setting/getting via the legacy functions. Or are you suggesting adding setters/getters and keep the previous setting/getting mechanism. 

Re thread-safety, it'd perhaps make customers think it is thread safe, but remember by definition TCP Direct is not thread safe; customers know if they want to ensure thread safety they'd need to do that by themselves or move to Onload. Are you suggesting that it'd worth making it thread safe for this specific case?

src/lib/zf/attr.c
```diff
@@ -300,13 +316,13 @@ int zf_attr_get_str(struct zf_attr* attr, const char* name, char** val)
...
-    if( *p != NULL ) {
+    if( strcmp(f->name, "log_file")==0 )
+      *val = strdup(zf_log_file_name);
+    else if( *p != NULL ) {
       *val = strdup(*p);
       if( *val == NULL )
         return -ENOMEM;
     }
-    else
-      *val = NULL;
```
jfeather-amd, line 309: Do we want to be deleting this else statement? I think it still makes sense in the case of some string field not being set. I expect a nice app would probably do something like:
```c
char* val = NULL;
int rc = zf_attr_get_str(..., &val)
if (rc < 0) return rc;
if (val) {
  // do something
}
```
But I worry specifically about the case where the consumer does not set `val = NULL`, especially as we were kind enough to do this for them previously.

src/lib/zf/doxygen/040_using.dox
```diff
@@ -113,8 +113,10 @@ production libraries:
   attribute, as described in the \ref attributes chapter.
 - As a convenience, the `-l` option to the `zf_debug` wrapper will set the
   \attrref{log_level} attribute to the specified value.
-- Changing the \attrref{log_level} attribute while the application is running
-  has no effect.
+- After the library initialization, the attributes attrref{log_level}, \attrref{log_format}
```
jfeather-amd, line 116:
```suggestion
- After the library initialization, the attributes \attrref{log_level}, \attrref{log_format}
```

src/lib/zf/log.c
```diff
@@ -143,6 +144,8 @@ int zf_log_redirect(const char* file)
 
   zf_log_stderr(); /* close any existing file */ 
   zf_log_file = f;
+  strcpy(zf_log_file_name, file);
```
jfeather-amd, line 147: Would you be able to change any `strcpy` calls to `strncpy`? Although I do wonder in this case if we should warn the user (or even error) on truncation.

jfeather-amd (commented):
Overall, this looks very solid, nice work! I have a couple minor bits of feedback, but feel free to dispute any if I've misunderstood their use cases.

Re @ligallag-amd's suggestion to have setters/getters, I imagine it could lead to simpler (and perhaps faster) code. Originally I didn't mind either approach, but the more I think about it, I think this separation is actually quite a good idea. This way it will also be clearer to users that the effect of, e.g., a public `zf_set_log_level` is broader than things set with `zf_attr_set_...`. All this being said though, I don't feel strongly enough to insist on this change blocking merge.

ligallag-amd:
@andresa-xilinx I was suggesting you add getters and setters to the public api and you get get rid of the get_attr/set_attr stuff. The static method you mention seems to only be used in a single file. I suppose doing this "locks in" the api in that logging is global and not per-stack - I don't know if we'd ever want to set logging per stack rather than globally. 

Indeed, customers trying to set log_level, log_file, etc would probably encounter an error if we disallowed getting/setting from `attr_set/get` but I wonder how bad that'd be. It did nothing in the past perhaps it could still do nothing.

The trouble with thread safety in this context is that setting/getting separate attribute interferes with each other as there is this global taint. Take for example attrs `a1` and `a2`, calling attr_set on `a1` for `log_file` and attr_get on `a2` for `log_file` will cause a rather nasty race despite those being entirely different attributes. This isn't the case for non-global attributes. Basically, we'd either need to be very explicit in our documentation that these attributes need to be synchronised on or we need to provide a way to synchronise on them.

src/lib/zf/attr.c
andresa-xilinx: agreed.

src/lib/zf/doxygen/040_using.dox
andresa-xilinx: done

src/lib/zf/log.c
andresa-xilinx: I've changed it to strncpy and issued a warning if truncated. Thanks!

src/lib/zf/zf.c
```diff
@@ -36,6 +42,13 @@ int zf_init(void)
     /* Failure here is non-fatal. */
   }
 
+ if( attr->log_file && strcmp(zf_log_file_name, attr->log_file) != 0 )
+    strcpy(zf_log_file_name, attr->log_file);
+ else if( attr->log_to_kmsg )
+    strcpy(zf_log_file_name, "/dev/kmsg");
+ else if( attr->log_file == NULL && zf_log_file_name[0] == '\0' )
+    strcpy(zf_log_file_name, "stderr");
```
jfeather-amd, line 50: I had a few comments about this block, so I'll merge them into a single comment as they somewhat depend on each other.

I think `attr->log_file` might also need to be handled carefully here to avoid overflows. Alternatively, can we drop this entire block? I think `zf_log_file_name` should be appropriately set by the above `zf_log_redirect`.

Should the `attr->log_to_kmsg` check come first here to match the above block, or is this the desired behaviour?

Should `"stderr"` here, be `"/dev/stderr"` instead? If we do drop this block, this should probably be the value we assign `zf_log_file_name` when we define it.

Also, if we do keep this block, I think the indentation of the `if`/`else if` lines are slightly off.

jfeather-amd (commented):
Thanks for addressing my review comments, @andresa-xilinx! Unfortunately, looking back over I have noticed some bits that might still need some work that I hadn't noticed before, sorry!

src/lib/zf/log.c
```diff
@@ -137,12 +138,19 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
 
 int zf_log_redirect(const char* file)
 {
-  FILE* f = fopen(file, "a");
+  strncpy(zf_log_file_name, file, sizeof(zf_log_file_name));
+  zf_log_file_name[ZF_LOG_FILE_NAME_SIZE-1] = '\0';
+
+  if( sizeof(file) > ZF_LOG_FILE_NAME_SIZE-1 )
```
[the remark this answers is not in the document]
andresa-xilinx: When you declare an array like `char a[10];`, the name `a` represents the whole array, not just a pointer to the first element (though it can also be used as such). `sizeof(a) == 10` in this case, and that was my intention in the code above. Moreover, the next line of code ensures that the destination string is always null-terminated (as `strncpy` does not guarantee so).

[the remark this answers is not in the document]
ligallag-amd: File is not of array type in this case, it's a pointer so `sizeof(file) ==8` in this case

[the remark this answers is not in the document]
jfeather-amd: Indeed, with a bit more testing, it also seems that changing `file` to be `char[]` still wouldn't be safe (as in this case I think it's implicitly cast to a `char*` when passed). See this [small snippet I used to test with godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:19,positionColumn:1,positionLineNumber:19,selectionStartColumn:1,selectionStartLineNumber:19,startColumn:1,startLineNumber:19),source:'%23include+%3Cstdio.h%3E%0A%0Asize_t+string_length(char*+str)+%7B%0A++return+sizeof(str)%3B%0A%7D%0A%0Asize_t+buf_length(char+str%5B%5D)+%7B%0A++return+sizeof(str)%3B%0A%7D%0A%0Aint+main(void)%0A%7B%0A++char+buf%5B512%5D+%3D+%22/dev/stderr%22%3B%0A%0A++printf(%22indirect+sizeof(char*)+%3D+%25ld%5Cn%22,+string_length(buf))%3B%0A++printf(%22direct+sizeof(char*)+%3D+%25ld%5Cn%22,+sizeof((char*)buf))%3B%0A++printf(%22indirect+sizeof(char%5B%5D)+%3D+%25ld%5Cn%22,+buf_length(buf))%3B%0A++printf(%22direct+sizeof(char%5B%5D)+%3D+%25ld%5Cn%22,+sizeof(buf))%3B%0A%0A++return+0%3B%0A%7D'),l:'5',n:'1',o:'C+source+%231',t:'0')),k:49.50101537568901,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:cclang_trunk,compilerName:'',compilerOutShown:'1',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',overrides:!(),runtimeTools:!(),source:1,stdinPanelShown:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+clang+(trunk)+(C,+Editor+%231)',t:'0'),(h:compiler,i:(compiler:cg141,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.1+(Editor+%231)',t:'0')),header:(),k:50.498984624311,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

src/lib/zf/doxygen/040_using.dox
```diff
@@ -113,8 +113,10 @@ production libraries:
   attribute, as described in the \ref attributes chapter.
 - As a convenience, the `-l` option to the `zf_debug` wrapper will set the
   \attrref{log_level} attribute to the specified value.
-- Changing the \attrref{log_level} attribute while the application is running
-  has no effect.
+- After the library initialization, the attributes \attrref{log_level}, \attrref{log_format}
+ attributes can be changed programmatically while the application is running using 
+ \attrref{zf_attr_set_int}.  Similarly, \attrref{log_file} can be modified
+ using \attrref{zf_attr_set_str}.
```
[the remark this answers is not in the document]
andresa-xilinx: yup, thanks for this.

src/lib/zf/log.c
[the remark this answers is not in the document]
andresa-xilinx: yes, you're correct. Sorry for missing this.

src/lib/zf/zf.c
andresa-xilinx: thanks, yes I think it is a good idea to drop most of the block. `zf_log_redirect` sets `zf_log_file_name` properly and the initial value of `zf_log_file_name` can be `/dev/stderr` when it is defined. I can't see how I can get rid of `attr->log_to_kmsg` case, which in turn made me realize `zf_log_replace_stderr` missedd setting `zf_log_file_name`.

src/lib/zf/attr.c
```diff
@@ -300,7 +316,9 @@ int zf_attr_get_str(struct zf_attr* attr, const char* name, char** val)
   switch( f->type ) {
   case zf_attr_type_str: {
     char** p = get_field_str(attr, f);
-    if( *p != NULL ) {
+    if( strcmp(f->name, "log_file")==0 )
```
jfeather-amd, line 319: Just spotted the use of `f->name` here where the other bits just use `name` directly. Does this mean the other places will now be case-sensitive (as I assume `f->name` is the lowercase name and `name` could reasonably be something like `LOG_LEVEL`)?

jfeather-amd (approved):
Good stuff! I'm quite happy with the state of this now and wouldn't mind merging as-is, though I do have one concern about the potential case-sensitivity of the `set_attr` functions (in these cases) that I hadn't spotted before. I would also like to see @ligallag-amd's opinion before merging this though, so I will re-request his review.

src/lib/zf/attr.c
pemberso-xilinx: It is broken as it currently stands. If zf_log_redirect fails then we have free(*p) but it is not set to NULL.
The original point is not addressed (can redirect but strdup fails).  Are you saying this doesn't matter because you don't use local attribute name?

src/lib/zf/attr.c
```diff
@@ -316,7 +316,9 @@ int zf_attr_get_str(struct zf_attr* attr, const char* name, char** val)
...
+      *val = strdup(zf_log_file_name);
```
pemberso-xilinx, line 320: should consider what happens if this strdup fails

src/lib/zf/log.c
```diff
@@ -135,14 +136,25 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
   va_end(va);
 }
 
+void zf_log_set_file_name(const char *file)
```
pemberso-xilinx, line 139: looks like only used in this file so should be static

src/lib/zf/log.c
```diff
@@ -135,14 +136,25 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
...
+{
+  strncpy(zf_log_file_name, file, sizeof(zf_log_file_name));
+  zf_log_file_name[ZF_LOG_FILE_NAME_SIZE-1] = '\0';
+
+  if( strnlen(file, ZF_LOG_FILE_NAME_SIZE) > ZF_LOG_FILE_NAME_SIZE-1 )
```
pemberso-xilinx, line 144: I would check the length is valid before trying to open or set name and error otherwise.  Don't think we should write to some truncated path that might not even be intended folder

src/lib/zf/log.c
```diff
@@ -135,14 +136,25 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
...
+
+  if( strnlen(file, ZF_LOG_FILE_NAME_SIZE) > ZF_LOG_FILE_NAME_SIZE-1 ) 
+    zf_log_stack_warn(NO_STACK, "WARNING: File name has been truncated as it is longer than %i characters\n", ZF_LOG_FILE_NAME_SIZE-1);
+}
+
 int zf_log_redirect(const char* file)
 {
-  FILE* f = fopen(file, "a");
+  zf_log_set_file_name(file);
+  FILE* f = fopen(zf_log_file_name, "a");
```
pemberso-xilinx, line 151: I would check opening before setting the name.  Opening could fail for several reason (e.g. permissions) so we wouldn't want the new name set.

pemberso-xilinx (changes requested):

src/lib/zf/attr.c
```diff
@@ -282,6 +283,7 @@ int zf_attr_set_str(struct zf_attr* attr, const char* name, const char* val)
       return -ENOMSG;
     }
     char** p = get_field_str(attr, f);
+    free(*p);
```
pemberso-xilinx, line 286: It is broken as it currently stands. If zf_log_redirect fails then we have free(*p) but it is not set to NULL.
The original point from previous commit is not addressed (can redirect but strdup fails). Are you saying this doesn't matter because you don't use local attribute name?

src/lib/zf/attr.c
andresa-xilinx: No, it means that I made an inconsistent (yet valid) change. I'll keep consistency and change it to `name`.

src/lib/zf/attr.c
andresa-xilinx: Correct. Thanks for spotting this  - I've set it to NULL. 

If strdup fails, not only will log use the global value (already set at the point you refer to), but also ENOMEM probably means there is a serious memory shortage hence I would not think an earlier evaluation would make a difference. Moreover, let suppose I do strdup first, if zf_log_redirect fails after strdup has already succeeded, the code must handle the deallocation of tmpdup which is more expensive (considering zf_log_redirect failing is more likely than a memory shortage).

src/lib/zf/log.c
andresa-xilinx: yes, thanks.

src/lib/zf/attr.c
andresa-xilinx: indeed, I've missed this one. Thanks.

src/lib/zf/log.c
andresa-xilinx: Agreed. Now it checks the length is valid before trying to open/set name and errors if not valid.

src/lib/zf/log.c
andresa-xilinx: `zf_log_set_file_name` now can return an error if `file` exceeds the max allowed length. It'll either fail because name is not of valid length or `fopen` is not successful. I prefer checking first whether the intended file name is a valid one, otherwise if it can be `fopen`ed and it is not a valid length, then I'd need to close/delete the empty file for nothing.

src/lib/zf/attr.c
jfeather-amd: My point was more that I feel like all of these should be `f->name` instead of just `name`. My understanding is that a user could call `zf_attr_get_str(attr, "LOG_FILE", &val)` and `zf_attr_info_find(name)` will match the appropriate attribute info (as it uses `strcasecmp` internally, so will match the lowercase string `"log_file"`). If you were to check `name` here in a case-sensitive way (`strcmp`) then something like `zf_attr_get_str(attr, "INTERFACE", &val)` would be valid, but `zf_attr_get_str(attr, "LOG_FILE", &val)` would not be.

andresa-xilinx: yes, I agree with you. I have overlooked `strcasecmp`. So all of these should be `f->name` to enforce matching the intended/appropriate attribute.

andresa-xilinx:
> The third commit seems to be doing more than just synchronizing as described. It looks like it is fixing comments from the second commit? I'd prefer clearer separation between the purpose of each commit. Possibly second and third commit should be merged?

yes, agreed.

andresa-xilinx:
>Basically, we'd either need to be very explicit in our documentation that these attributes need to be synchronised on or we need to provide a way to synchronise on them.

@ligallag-amd I have documented as a known issue that global attributes need to be synchronized by the app:

https://xilinx.lightning.force.com/lightning/r/Knowledge__kav/ka04U000000y4I0QAI/view

Would this be enough?

src/lib/zf/doxygen/040_using.dox
```diff
@@ -113,9 +113,10 @@ production libraries:
   attribute, as described in the \ref attributes chapter.
 - As a convenience, the `-l` option to the `zf_debug` wrapper will set the
   \attrref{log_level} attribute to the specified value.
-- After the library initialization, the attributes attrref{log_level}, \attrref{log_format}
- attributes can be changed programmatically while the application is running using 
- \attrref{zf_attr_set_int}.
+- After library initialization, the \attrref{log_level} and \attrref{log_format}
+  attributes can be changed programmatically while the application is running using 
+  \attrref{zf_attr_set_int}.  Similarly, \attrref{log_file} can be modified
+  using \attrref{zf_attr_set_str}.
```
ligallag-amd, line 119: I think it's worth adding a note that these attributes are global. e.g "\attrref{log_level} and \attrref{log_format} are global attributes and may be modified by anythread. Therefore, any users setting these levels from both threads must ensure thread safety."

ligallag-amd (commented):
I know it's pedantic, but your second commit message isn't formatted properly. A line may be at most 72 characters. I also think you can remove: "Also, synchronizing global and local attributes when instantiating multiple attribute sets." - I don't believe you are doing any sort of synchronisation (no locking).

I don't understand: "Now there is a single global initialization, and subsequent attribute sets instantiation won't overwrite values of a global attribute.". Where have you done this? To me, it looks like other attribute sets are capable of modifying the log level?

andresa-xilinx:
> I know it's pedantic, but your second commit message isn't formatted properly. A line may be at most 72 characters. I also think you can remove: "Also, synchronizing global and local attributes when instantiating multiple attribute sets." - I don't believe you are doing any sort of synchronisation (no locking).

agreed. 

> 
> I don't understand: "Now there is a single global initialization, and subsequent attribute sets instantiation won't overwrite values of a global attribute.". Where have you done this? To me, it looks like other attribute sets are capable of modifying the log level?

I meant before the initial values set at `zf_init()` were overwritten by `zf_attr_alloc(...)`:

https://github.com/Xilinx-CNS/tcpdirect/pull/8/files#diff-9cb43605a7d0ab39a4c5c26d08f3e8de261d4ff17cb674059999a4b69c0d2663L140

By instantiation I meant during allocation of new set of attributes (e.g., see the unit test below), which now respect the current global values.

https://github.com/Xilinx-CNS/tcpdirect/pull/8/files#diff-a18c5e88dd663f03845ff5d519eaf2becfed1d9a4a78d05760bee14f250948b1R211

src/lib/zf/doxygen/040_using.dox
andresa-xilinx: agreed. Thank you.

pemberso-xilinx (changes requested):
Code looks ok.  Remaining comments are on organization of commits and making sure commit message match the changes in the commit.

I'd also like to see some test output where changing log_file fails.  E.g. try setting it to a path that doesn't exist or you don't have permission to write to.

src/lib/zf/doxygen/040_using.dox
```diff
@@ -113,9 +113,13 @@ production libraries:
...
+- \attrref{log_level} and \attrref{log_format} are global attributes and may be 
+  modified by any thread. Therefore, any users setting these levels from both 
+  threads must ensure thread safety.
```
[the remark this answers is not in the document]
ligallag-amd: Probably best to just squash the commits.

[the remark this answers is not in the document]
andresa-xilinx: done, both commits are now squashed.

src/tests/zf_unit/zfattrs.c
```diff
@@ -161,19 +173,94 @@ static void test(struct zf_stack* stack, struct zf_attr* attr,
...
+  if(  tmp_log_dir == NULL )
+    return errno;
+
+  char *non_existent_dir = mkdtemp(log_template_2);
+  if( non_existent_dir == NULL )
+    return errno; 
+
+  int rc = rmdir(non_existent_dir);
+  if( rc < 0 )
+    return rc;
```
krishd-amd, line 189: Why is this creating a tempdir and then immediately deleting it (non_existent_dir)?

andresa-xilinx: Because I need to test a non-existent path in:

https://github.com/Xilinx-CNS/tcpdirect/pull/8/files/43dbf5afc8292db609ea74d060aade215e096616#diff-a18c5e88dd663f03845ff5d519eaf2becfed1d9a4a78d05760bee14f250948b1R209

Is there another way you'd prefer ensuring the tested path does not actually exist?

krishd-amd: I think it would be a little cleaner to avoid doing IO if we can avoid it.

Could we instead generate a temporary file name and stat it to see it doesn't exist?

andresa-xilinx: TTBOMK there is not a standard C function that returns only the name of a non-existent file without creating it. I didn't think it worth introducing the complexity of adding a function which generates UID when it's only used once, and for this specific test. Moreover, after writing the UID generator function we'd perhaps start discussing about the probability of collision for that UID. Delegating the responsibility to libc seemed the safest choice. 

I understand that if we had the name only we would avoid a couple of write operation in the disc which will be part of the test anyway as we're writing long logs with 100's of lines (as test process) to the disc. 

Having said that, let me know if you'd be OK with this exception or rather want me to write the UID function; but would appreciate further clarification on why that'd make it cleaner.

pemberso-xilinx: I don't think you need a C function or uid generator.  "/path/to/nonexistent/file" is very unlikely to exist and you can stat() it as Krish suggested to be sure.

andresa-xilinx: Yes, but pedantically speaking it is less safe than mkdtemp and I don't see the problem in terms of the I/O cost. Anyway, as that's the consensus, I will change it to '/path/to/nonexistent/file"

krishd-amd (approved):

src/lib/zf/log.c
```diff
@@ -135,14 +136,44 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
...
+static int zf_log_set_file_name(const char *file)
+{
+  if( strnlen(file, ZF_LOG_FILE_NAME_SIZE) > ZF_LOG_FILE_NAME_SIZE-1 ) {
+    zf_log_stack_err(NO_STACK, "%s: File name is longer than %i characters: %s\n",
+                     __func__, ZF_LOG_FILE_NAME_SIZE-1, strerror(-EINVAL));
+    return -EINVAL;
+  }
+
+  strncpy(zf_log_file_name, file, sizeof(zf_log_file_name));
+  zf_log_file_name[ZF_LOG_FILE_NAME_SIZE-1] = '\0';
```
ligallag-amd, line 148:
```suggestion

```
strncpy will null terminate the destination so there's no need for this.

src/lib/zf/log.c
```diff
@@ -135,14 +136,44 @@ void zf_log(struct zf_stack* st, const char* fmt, ...)
...
 
+  int rc = zf_log_set_file_name(file);
+  if( rc < 0 ) {
+    free(previous_file_name);
+    return rc;
+  }
+
+  FILE* f = fopen(zf_log_file_name, "a");
+  if( ! f ) {
+    strncpy(zf_log_file_name, previous_file_name, sizeof(zf_log_file_name));
```
ligallag-amd, line 167:
```suggestion
  if( !f ) {
    zf_log_set_file_name(previous_file_name)
```

Prefer function reuse despite redundant checks

ligallag-amd (approved):
apply suggested changes, test and squash.

src/lib/zf/log.c
andresa-xilinx: thanks, and the correct size is checked just above, so it's redundant.

andresa-xilinx: included and tested.

PR #10, opened 2024-04-23
cleanup: CI tweaks

jfeather-amd:
This commit makes a few minor changes to the tests:
- allow the Jenkins tests to fail and continue with the rest of the pipeline;
- run through all tests in GitHub actions, delaying failure to the end;
- log failed tests to the perform_build GitHub action stage summary;
- move the `testmisc` target to Jenkins for now, until it is fixed;
- flush stdout in assertions to avoid `SIGABRT` losing test output;
- switch to using the public onload repository for GitHub actions;
- improve skipping packetdrill tests for GitHub actions; and
- remove unused valgrind test wrappers.

### Testing Done
- GitHub actions run [with a failure](https://github.com/jfeather-amd/tcpdirect/actions/runs/8802425878) (before moving testmisc to Jenkins), scroll down on the linked page for the failure summary
- Passing GitHub action tick on this commit ([see here for the full run](https://github.com/jfeather-amd/tcpdirect/actions/runs/8803892680))
- Jenkins pipeline run, check my personal pipeline or message me for this

Note that the runs skip `packetdrill` because I didn't set up the necessary keys.

### DIFF
--- a/src/tests/zf_unit/netns.sh
+++ b/src/tests/zf_unit/netns.sh
@@ -21,4 +21,36 @@ echo 4096 > /proc/sys/vm/nr_hugepages
 ifconfig lo up
 shift
 
-exec "$@"
+output="$(timeout ${TEST_TIME_OUT:-120} $@)"
+rc="$?"
+echo "$output"
+if [ "$rc" -ne 0 -a -n "$UT_OUTPUT" ]; then
+  if [ ! -f "$UT_OUTPUT" ]; then
+    output_dir="$(dirname $UT_OUTPUT)"
+    if [ -n "$output_dir" -a ! -d "$output_dir" ]; then
+      mkdir -p "$output_dir"
+    fi
+    touch "$UT_OUTPUT"
+  fi
+
+  echo "# Failed \`$(basename $@)\` in \`$TEST_TARGET\`" >> "$UT_OUTPUT"
+  if [ "$rc" -eq 124 ]; then
+    echo "Test timed out (rc=$rc)." >> "$UT_OUTPUT"
+  else
+    echo "Test returned $rc." >> "$UT_OUTPUT"
+  fi
+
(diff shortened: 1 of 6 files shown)


krishd-amd (approved):
looks fine to me

krishd-amd:
do you perhaps also know why the action is not being triggered here?

jfeather-amd:
> do you perhaps also know why the action is not being triggered here?
> 
> I think we need to edit hte top to also trigger on pr.

Yes, I expect you're right, checking other PRs seems to show that it hasn't triggered on any of them.

> Also, would you mind doing a quick PR to my tcpdirect fork too? I think I have it such that external PRs need our approval before running the pipeline.

Done, see krishd-amd/tcpdirect#9.

jfeather-amd:
Latest push adds:
- run GitHub actions when pull requests are created, in addition to pushes;

src/tests/zf_unit/netns.sh
```diff
@@ -21,4 +21,36 @@ echo 4096 > /proc/sys/vm/nr_hugepages
...
+  fi
+
+  echo "# Failed \`$(basename $@)\` in \`$TEST_TARGET\`" >> "$UT_OUTPUT"
+  if [ "$rc" -eq 124 ]; then
+    echo "Test timed out (rc=$rc)." >> "$UT_OUTPUT"
+  else
+    echo "Test returned $rc." >> "$UT_OUTPUT"
+  fi
+
+  echo "## Suggested Reproducer" >> "$UT_OUTPUT"
```
ivatet-amd, line 43: I wanted to suggest using HereDocs, but a few moments late for a party! (Either way is OK.)

jfeather-amd: From a brief look, that probably would give a much nicer-to-read bit of code here! I created this bit with the intention that it could be updated with a nicer markdown summary, so perhaps something to follow up on when this code is next changed!

PR #11, opened 2024-04-24
cleanup: use apt-get rather than apt in github actions

krishd-amd:
Problem: 

https://github.com/Xilinx-CNS/tcpdirect/actions/runs/8814928979/job/24195841663

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.


jfeather-amd (approved):
Seems reasonable.

PR #14, opened 2024-05-24
Ensure zft_handle laddr is initialized on reuse

pemberso-xilinx:
Previously when a zft_handle was re-used after being free'd the laddr was not reinitialized.

First commit modifies zfzockfree test to demonstrate failure.
Second commit is proposed fix.

Failing test: [on_15816_test_fail.txt](https://github.com/Xilinx-CNS/tcpdirect/files/15433813/on_15816_test_fail.txt)
See
```
ok 449 - Freed TCP zocket 61
ok 450 - Freed TCP zocket 62
ok 451 - Freed TCP zocket 63
not ok 452 - Reallocated TCP zocket 0
#   Failed test 'Reallocated TCP zocket 0'
#   at src/tests/zf_unit//zfzockfree.c line 246.
#     -16
#         ==
#     0
```

and

```
ok 578 - Freed TCP zocket 62 again
ok 579 - Freed TCP zocket 63 again
Bail out!  Failed to allocate TCP handle 0 (rc = -16)
```

Passing test: [Uploading on_15816_test_pass.txt…]()

### DIFF
--- a/src/lib/zf/tcp.c
+++ b/src/lib/zf/tcp.c
@@ -64,6 +64,7 @@ static int __zft_handle_free(zf_stack* stack, zf_tcp* tcp)
 
   zf_stack_tcp_to_res(stack, tcp, &rx_res);
   zfrr_release_port(stack, rx_res);
+  memset(&tcp->laddr, 0, sizeof(tcp->laddr));
 
   if( tcp->eof_pkt != PKT_INVALID ) {
     zf_pool_free_pkt(&stack->pool, tcp->eof_pkt);
--- a/src/tests/zf_unit/zfzockfree.c
+++ b/src/tests/zf_unit/zfzockfree.c
@@ -128,9 +128,16 @@ static int create_zocket(struct zf_stack* stack, struct zf_attr* attr,
                          struct zft** tcp_out)
 {
   struct zft_handle* tcp_handle;
+  struct sockaddr_in laddr;
+  socklen_t len = sizeof(laddr);
+
   int rc = zft_alloc(stack, attr, &tcp_handle);
   if( rc < 0 )
     return rc;
+
+  zft_handle_getname(tcp_handle, (struct sockaddr *)&laddr, &len);
+  if( laddr.sin_port != 0 )
+    return -EBUSY;
 
   /* Do a loopback connect.  There's nothing listening, but we don't care: we
    * just want to install the filters. */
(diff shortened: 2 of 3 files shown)


src/lib/zf/tcp.c
```diff
@@ -64,6 +64,7 @@ static int __zft_handle_free(zf_stack* stack, zf_tcp* tcp)
 
   zf_stack_tcp_to_res(stack, tcp, &rx_res);
   zfrr_release_port(stack, rx_res);
+  memset(&tcp->laddr, 0, sizeof(tcp->laddr));
```
jfeather-amd, line 67: I think this might be superfluous if we're also choosing to also zero this in `zf_tcp_release` as we call that function at the bottom of this function (with no chance to exit early without calling it). Given that it's just some extra work at free time though, I don't think it should matter much if you don't want to change it.

src/tests/zf_unit/zfzockfree.c
```diff
@@ -128,10 +128,17 @@ static int create_zocket(struct zf_stack* stack, struct zf_attr* attr,
...
+  struct sockaddr_in laddr;
+  socklen_t len = sizeof(laddr);
+
   int rc = zft_alloc(stack, attr, &tcp_handle);
   if( rc < 0 )
     return rc;
 
+  zft_handle_getname(tcp_handle, (struct sockaddr *)&laddr, &len);
+  if( laddr.sin_port != 0 )
+    return -EBUSY;
```
jfeather-amd, line 140: Why is it that `-EBUSY` was chosen here? Is it possible for a handle to still be in use (with some delayed shutdown) after we free it, so we end up here after trying to reuse it after alloc? Or is it just a useful assertion for this specific issue you're fixing? (Just curious, not saying `-EBUSY` should be something else)

jfeather-amd (approved):
Looks good! Just a couple of comments, but happy for this to be merged as-is.

krishd-amd (approved):

src/tests/zf_unit/zfzockfree.c
pemberso-xilinx: It was the second point (a useful assertion) - needed a suitable error to indicate test failure.  I wasn't sure what error was best but thought EBUSY indicated the handle was in some way still in use by still having a local address set.

src/lib/zf/tcp.c
pemberso-xilinx: Good point!  I won't update now as it has been merged but agree it is probably superfluous.

PR #15, opened 2024-05-25
ON-15689: use public accessor macros to read ef_vi tx ts data

abower-amd:
Fixes problem whereby TCPDirect treated the bottom two bits of the timestamp nanosecond part as sync flags when these have now moved to a separate field in the event structure.

### DIFF
--- a/src/include/zf_internal/tx_reports.h
+++ b/src/include/zf_internal/tx_reports.h
@@ -224,10 +224,16 @@ complete(struct queue* q, unsigned z, bool tcp, ef_event* ev)
   zf_assert(node->zock == zock_id(z, tcp));
   zock->pending = node->zock_next;
 
-  node->report.timestamp = {ev->tx_timestamp.ts_sec, ev->tx_timestamp.ts_nsec};
+  node->report.timestamp = {
+    .tv_sec  = EF_EVENT_TX_WITH_TIMESTAMP_SEC(*ev),
+    .tv_nsec = EF_EVENT_TX_WITH_TIMESTAMP_NSEC(*ev),
+  };
+
+  /* While ef_vi and TCPDirect have the same sync flags layout they can be
+   * copied without translation. */
   static_assert(EF_VI_SYNC_FLAG_CLOCK_SET == ZF_PKT_REPORT_CLOCK_SET);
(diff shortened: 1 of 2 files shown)


jfeather-amd:
This looks like a reasonable change, but I wonder if we might want to update the [timestamping unit test](https://github.com/Xilinx-CNS/tcpdirect/blob/85cf55a048ea2ed25dd91b8e3dbcc6ad9b8638f1/src/tests/zf_unit/zftimestamping.c)  to also include things like flags and make sure they're set. Perhaps these could be set to `i & 3` for the sake of [expected_report](https://github.com/Xilinx-CNS/tcpdirect/blob/85cf55a048ea2ed25dd91b8e3dbcc6ad9b8638f1/src/tests/zf_unit/zftimestamping.c#L36)/[check_report](https://github.com/Xilinx-CNS/tcpdirect/blob/85cf55a048ea2ed25dd91b8e3dbcc6ad9b8638f1/src/tests/zf_unit/zftimestamping.c#L44)?

jfeather-amd (approved):
Previous comment was supposed to be an approval - forgot to change the selected button! I'm currently looking at the tests that aren't too consistent and found that this patch seems to fix the txtimestamping tests too, although it requires the following (quick fix) to resolve a build error:
```diff
diff --git a/src/tests/zf_unit/zftimestamping.c b/src/tests/zf_unit/zftimestamping.c
index fd3bfa22..e7ced36e 100644
--- a/src/tests/zf_unit/zftimestamping.c
+++ b/src/tests/zf_unit/zftimestamping.c
@@ -83,6 +83,7 @@ static void tx_report_batch(struct zf_tx_reports::queue* reports, int capacity,
         ev.tx_timestamp.type = EF_EVENT_TYPE_TX_WITH_TIMESTAMP;
         ev.tx_timestamp.ts_sec = r.timestamp.tv_sec;
         ev.tx_timestamp.ts_nsec = r.timestamp.tv_nsec;
+        ev.tx_timestamp.ts_flags = 0;
         zf_tx_reports::complete(reports, z/2, z%2, &ev);
       }
     }

```
I imagine this didn't come up in testing as this test is hidden away in `make test_jenkins`, as I'm in the area I'd be happy to add the suggested testing I mention in my previous comment - although I think it'd be best to make sure this compiles with `make test_jenkins` too!

PR #21, opened 2024-06-04
cleanup: cherry-pick test/CI changes to v8_1

jfeather-amd:
Cherry pick a bunch of (mainly CI/test) changes to v8_1.

The top commit (e367fb453f0e6d5a111d9f67c0f3ead94e03d4d4) is likely to be of interest as that is the only real change here (that isn't being cherry-picked). Because there is no pairing `v8_1` branch in the Xilinx-CNS/onload repo, we can't directly link to that as we do with `master`. Until a better solution is unlocked, building against the latest OpenOnload version is probably to best we can do (it might be nice to have a `v8_x` branch to avoid needing to update code here).

Cherry-picks:
- ~1e5c1f594439ca838fbd1a9165832496c7ca0190 "Cleanup: Update Jenkins node names"~
- 39f118e0850b785f91fda7ce4eee3ff4421915e0 "ON-15255: Introduce build and unit test checks as github actions"
- 45a1c9fbf6f841e98da0c0a9523acfa409959600 "ON-15231: Simplify Makefile.inc by removing junit formatting options"
- fef3a44b98c8e527fefa55aba84a0cee2addee01 "ON-15255: Reduce Jenkins system tests to a smaller subset of unit tests"
- 8f0ee663016cfb30ab8b5e67faaf8118c6ca1f99 "Cleanup: fix out-of-bounds efct_shm access in zf_emu"
- 105e1230eeee94515ec2cda93738371d503ca566 "cleanup: CI tweaks"
- 00851457dd46838a867b18a6a95f2357b732841d "cleanup: use apt-get rather than apt in github actions"
- 378d0fbe6ef1cb14bb65b6da5d32af9251e4261d "ON-15724: maintain zf_emu mib maximum hwports"
- c31b2023231f26ce12b3c9d1b4419a471ec23343 "ON-15724: ensure Jenkins-only tests have proper attributes"
- eaa2ff607f74925eaeb79b5be63fcd2d8500a7f3 "ON-15724: ensure test scripts log properly"
- cfc312b56f1750e062e6241d4457151e183b9487 "ON-15724: merge pretest.sh into netns.sh"
- a61dbe41cd57d8593c7c3870cd99ec0809516378 "ON-15724: cleanup zf emu shared memory in test scripts"
- ace0b69086020858c85316ff0da7128815b6429b "ON-15724: change method of reading fd in zfstackdump test"

### Testing Done
- GitHub actions run [looks happy](https://github.com/jfeather-amd/tcpdirect/actions/runs/9365168957)
- Subsequent merge to master looks reasonable ([see my fork](https://github.com/jfeather-amd/tcpdirect/tree/master)), notably this doesn't bring the `release_tag` from versions.yaml along with it as `master` probably shouldn't have a release.
- Jenkins looks happy-ish (see my personal pipelines), although it seems zfsend.sh is failing consistently on testscript_x3 - I expect we might need to loosen the timeout here as it runs fine on my machine and it probably does legitemately take that long to init zf_emu sometimes.
    - As an aside, it would be nice to also add a summary step to the pipeline so we get some error reporting - e.g., add `UT_OUTPUT` to the test invocation then another step to `cat` the contents of the summary file.


tcrawley-xilinx:
Do we need `Cleanup: Update Jenkins node names`? I worry that we might miss build/test coverage on RHEL7

jfeather-amd:
> Do we need `Cleanup: Update Jenkins node names`? I worry that we might miss build/test coverage on RHEL7

Perhaps not, if that's why it was done then, I couldn't quite remember which bug was linked to that change so included it because it seemed inoffensive (though I was clearly wrong)! I'll drop that commit from this PR.

jfeather-amd:
This set of commits still looks to be valid for cherry picking, the only new commits [between v8_1 and master](https://github.com/Xilinx-CNS/tcpdirect/compare/v8_1...master) (see those starting at ON-15255, da134ec) appear to be valid "master-only" commits.

sianj-xilinx:
Can you explain more why we have to use a release tag for 8.1 builds rather than, say, last_known_good?

ligallag-amd (approved):

krishd-amd (approved):

jfeather-amd:
> Can you explain more why we have to use a release tag for 8.1 builds rather than, say, last_known_good?

With GitHub actions, we're restricted to using [Xilinx-CNS/onload](https://github.com/Xilinx-CNS/onload) if we don't want to add repo secrets for pulling private/internal repos. This means we only have access to the onload `master` branch and any tags that were created for the release of OpenOnload (which are `v8.1.[0..3]` as of writing). This may change in the future, but it appears that this is the best we can do for tcpdirect v8_1 actions for now. Notably though, Jenkins will still build against the last known good branches (now in the form of tarballs), so perhaps it will be worth still running all tests on Jenkins for v8_1 to get a full view with the latest 8.1 changes.

sianj-xilinx (approved):
This looks like the best we can do at the moment, but I think it's only a temporary solution as it will break as soon as we make any incompatible changes to onload-8.1. Let's merge this, but look at ways we can enable access to the actual branch. @krishd-amd has some ideas about how we can do this.

PR #23, opened 2024-07-01
ON-15163: Update to use generic warming functions

ibeecraft-amd:
Change has been made to deprecate the architecture specific and transmit method specific warming functions. Generic functions for enabling or disabling transmit warming functionality should be used instead.

This commit updates TCPDirect to use the generic ef_vi functions.

Testing was done with zftcppingpong_sleep

Below is perf trace showing new generic methods used on X3 for transmit warming.
tcpdirect_warming

Test was also ran on X2.

### DIFF
--- a/src/lib/zf/tx_warm.c
+++ b/src/lib/zf/tx_warm.c
@@ -14,22 +14,16 @@ int enable_tx_warm(struct zf_tx* tx, zf_tx_warm_state* state)
   ef_vi* vi = zf_stack_nic_tx_vi(st_nic);
   zf_assert_nflags(st->flags, ZF_STACK_FLAG_TRANSMIT_WARM_ENABLED);
   zf_log_stack_trace(st, "%s: TX warm enabled\n", __func__);
-  /* TODO better way to distuinguish X2 vs X3 style CTPIO */
-  if( vi->nic_type.arch == EF_VI_ARCH_EFCT ) {
-    efct_vi_start_transmit_warm(vi);
+  char* ctpio_warm_buf = NULL;
+  state->ctpio_warm_buf_id = PKT_INVALID;
+  if( vi->vi_flags & EF_VI_TX_CTPIO && vi->nic_type.arch != EF_VI_ARCH_EFCT ) {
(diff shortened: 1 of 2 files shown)


src/lib/zf/tx_warm.c
```diff
@@ -42,15 +36,9 @@ void disable_tx_warm(struct zf_tx* tx, zf_tx_warm_state* state)
...
-    efct_vi_stop_transmit_warm(vi);
-  }
-  else {
-    ef_vi_stop_transmit_warm(vi, &state->ef_vi_state[tx->path.nicno]);
-    if( state->ctpio_warm_buf_id != PKT_INVALID )
-      zf_pool_free_pkt(&st->pool, state->ctpio_warm_buf_id);
-  }
+  ef_vi_stop_transmit_warm(vi, &state->ef_vi_state[tx->path.nicno]);
+  if( state->ctpio_warm_buf_id != PKT_INVALID )
+    zf_pool_free_pkt(&st->pool, state->ctpio_warm_buf_id);
```
ligallag-amd, line 41: Is it possible for this to be called on efct? If so, will this do anything? Before this patch this was not called and I'm wondering if there was a reason for that.

jfeather-amd: It looks safe, `state->ctpio_warm_buf_id = PKT_INVALID;` unless `vi->vi_flags & EF_VI_TX_CTPIO && vi->nic_type.arch != EF_VI_ARCH_EFCT`, in which case it's assigned as a result of `zft_alloc_pkt`.

ibeecraft-amd: The code in the enable function is writing to avoid calling zf_pool_free_pkt on efct. If you called free on a pkt not assigned that would cause issues but on efct the buf_id will be PKT_INVALID.  The ef_vi_state is ignored on EFCT so makes no difference.

jfeather-amd (approved):
Looks good to me, although I seem to be the culprit of the failing build!

ligallag-amd (approved):

jfeather-amd:
I'll merge this now (after #24) so you get a nice green tick on your commit rather than failing because of me!

PR #25, opened 2024-07-08
ON-15875: update zf_mkdist (and _build) to work in a wider context

jfeather-amd:
These would previously require a very specific setup, needing some internal build server with some assumptions made about CC as a result of this. This is no longer necessary, and has started causing issues with compilation errors due the addition of `memfd_create` in onload which isn't supported by the kernel on these build machines.

### Testing Done
The following sequence of commands now works on my local Ubuntu machine without access to these servers:
```
$ ls
onload tcpdirect
$ cd onload
$ ./scripts/onload_mkdist --version 9.0.0
$ cd ../tcpdirect
$ ./scripts/zf_mkdist --shim --version 9.0.0 "$PWD/../onload/onload-9.0.0.tgz"
```
I've tested the above with both master and v8_1. Further internal testing is documented on the associated ticket.


tcrawley-xilinx (commented):
Changes look good if we are going to stick with `zf_mkdist` and `zf_mkdist_build` as they are. I would like to see what a unified script would look like now that we don't have to deal with a build server getting in the way.

jfeather-amd:
> I would like to see what a unified script would look like

This seems like it could be a reasonable thing to do; a [quick search](https://github.com/search?q=org%3AXilinx-CNS+%2Fzf_mkdist%28_build%29%3F%2F&type=code) shows that our Jenkins pipeline is the only direct consumer of `zf_mkdist_build`:
https://github.com/Xilinx-CNS/tcpdirect/blob/7d4498c5e7fb1c1e2c40fed0fda218e81124e883/ci/test.groovy#L179-L191

Although the CI pipeline may become a bit more obtuse if we have to `onload_mkdist` first, so perhaps I'll see if it's too tricky to allow either an onload tarball or an onload source tree.

abower-amd:
> Although the CI pipeline may become a bit more obtuse if we have to `onload_mkdist` first, so perhaps I'll see if it's too tricky to allow either an onload tarball or an onload source tree.

I think we want to move toward RPM build consuming Onload (S)RPM, TGZ build consuming Onload TGZ, etc.

jfeather-amd:
Latest push moves to use the onload tarball instead of a source tree clone, and merges zf_mkdist_build into zf_mkdist.

tcrawley-xilinx (approved):

PR #26, opened 2024-07-15
ON-15255: use ef_vi packet in timestamp calls

jfeather-amd:
Commit da134ec0d456 (ON-15255: use new unified ef_vi timestamping API, 2024-07-01) was an improper patch, and used to wrong packet reference; instead of referencing the ef_vi rxpkt it instead tried to reference the TCPDirect packet store. This change references the correct packet to correctly extract a timestamp.

This change also teaches zf_emu to expect sub-nanosecond resolution timestamps, rather than the previous clobbering of the lower-order bits in the nanoseconds field.

### Testing
zf_emu is happy running x3 tests again, although why they were always happy on github is a mystery to me.

### DIFF
--- a/src/lib/zf/private/reactor.c
+++ b/src/lib/zf/private/reactor.c
@@ -151,16 +151,18 @@ zf_reactor_handle_rx_ref_discard(struct zf_stack* st, int nic, ef_vi *vi,
 }
 
 
+/* This function places a packet prefix at `dest` for a received EFCT packet.
+ * The `efpkt` parameter takes the value retrieved from `efct_vi_rxpkt_get` */
 ZF_HOT static void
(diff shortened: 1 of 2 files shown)


src/lib/zf/private/reactor.c
```diff
@@ -152,15 +152,15 @@ zf_reactor_handle_rx_ref_discard(struct zf_stack* st, int nic, ef_vi *vi,
 
 
 ZF_HOT static void
```
sianj-xilinx, line 154: Perhaps document that efpkt is a packet pointer retrieved through `efct_vi_rxpkt_get()`?

sianj-xilinx (approved):

ligallag-amd (approved):

jfeather-amd:
Addressed review comment, will wait for the actions to finish running then will merge.

PR #28, opened 2024-07-19
ON-15895: handle TX completions for in-flight retransmitted packets

jfeather-amd:
If we have not yet processed a TX completion for a given packet and the RTO timer expires, we will retransmit it via PIO or by copying it to a new buffer. When adding TX timestamping, these packets would be marked to be freed when the TX completion was processed. However, we would still prepare a TX timestamp report for it, thus expecting the report to be completed.

The problem with this is that the flag used on retransmit to request the packet gets freed immediately, `ZF_REQ_ID_PROTO_TCP_FREE`, would indeed free the packet immediately without informing the TX report system of this TX completion event. The next sent packet would then populate the TX report for this retransmission with the timestamp in its TX completion event. Subsequent packets would also experience this off-by-one failure.

This failure mode can be triggered by sending some data, then not performing the reactor until the RTO timer must have expired, at which point performing the reactor again would handle the timer expiration before checking for the probably-already-present TX completion, meaning we get this specific flavour of retransmit.

To fix this, the send flag has been updated so that we process the TX completion, to be consistent with other retransmissions which get TX reports. This also required a minor update to correctly label this report as a retransmission when preparing the report.

### Testing Done
The reproducer for this issue created by @ibeecraft-amd now observes all of the expected TX reports, including the retransmitted packet.

<details><summary>Failing Log (pre-patch)</summary>
<p>

```
Sending Payload: 64
Count of timestamps:2
Timestamps: start:0, bytes:0, flags:4001, seconds:1719325773 nsecs:209034425 
Timestamps: start:0, bytes:64, flags:1, seconds:1719325773 nsecs:209095661 
Sending Payload: 65
#### LONG SLEEP HAPPENS HERE (so we don't handle TX completion of above) ####
Count of timestamps:1
Timestamps: start:0, bytes:64, flags:2001, seconds:1719325775 nsecs:209259357 
Sending Payload: 66
Count of timestamps:1
Timestamps: start:64, bytes:65, flags:1, seconds:1719325775 nsecs:209308825 
Sending Payload: 67
Count of timestamps:1
Timestamps: start:129, bytes:66, flags:1, seconds:1719325775 nsecs:209344369 
Sending Payload: 68
Count of timestamps:1
Timestamps: start:195, bytes:67, flags:1, seconds:1719325775 nsecs:209380329 
Sending Payload: 69
Count of timestamps:1
Timestamps: start:262, bytes:68, flags:1, seconds:1719325775 nsecs:209415601 
Sending Payload: 70
Count of timestamps:1
Timestamps: start:330, bytes:69, flags:1, seconds:1719325775 nsecs:209450617 
Sending Payload: 71
Count of timestamps:1
Timestamps: start:399, bytes:70, flags:1, seconds:1719325775 nsecs:209487545 
Sending Payload: 72
Count of timestamps:1
Timestamps: start:469, bytes:71, flags:1, seconds:1719325775 nsecs:209522617 
Sending Payload: 73
Count of timestamps:1
Timestamps: start:540, bytes:72, flags:1, seconds:1719325775 nsecs:209561225 
```

</p>
</details> 

<details><summary>Passing Log (post-patch)</summary>
<p>

```
Sending Payload: 64
Count of timestamps:2
Timestamps: start:0, bytes:0, flags:4001, seconds:1721381680 nsecs:401417725
Timestamps: start:0, bytes:64, flags:1, seconds:1721381680 nsecs:401690641
Sending Payload: 65
#### LONG SLEEP HAPPENS HERE ####
Count of timestamps:1
Timestamps: start:64, bytes:65, flags:1, seconds:1721381680 nsecs:401902045
Count of timestamps:1
#### NOTE flags & 0x2000 DENOTES RETRANSMIT ####
Timestamps: start:64, bytes:65, flags:2001, seconds:1721381700 nsecs:402331909
Sending Payload: 66
Count of timestamps:1
Timestamps: start:129, bytes:66, flags:1, seconds:1721381700 nsecs:402559097
Sending Payload: 67
Count of timestamps:1
Timestamps: start:195, bytes:67, flags:1, seconds:1721381700 nsecs:402742185
Sending Payload: 68
Count of timestamps:1
Timestamps: start:262, bytes:68, flags:1, seconds:1721381700 nsecs:402919001
Sending Payload: 69
Count of timestamps:1
Timestamps: start:330, bytes:69, flags:1, seconds:1721381700 nsecs:403111065
Sending Payload: 70
Count of timestamps:1
Timestamps: start:399, bytes:70, flags:1, seconds:1721381700 nsecs:403291297
Sending Payload: 71
Count of timestamps:1
Timestamps: start:469, bytes:71, flags:1, seconds:1721381700 nsecs:403470269
Sending Payload: 72
Count of timestamps:1
Timestamps: start:540, bytes:72, flags:1, seconds:1721381700 nsecs:403637493
Sending Payload: 73
Count of timestamps:1
Timestamps: start:612, bytes:73, flags:1, seconds:1721381700 nsecs:403808757
```

</p>
</details>

### DIFF
--- a/src/lib/zf/tcp_out.c
+++ b/src/lib/zf/tcp_out.c
@@ -903,6 +903,20 @@ int tcp_output(struct zf_tcp* tcp)
                                     eff_right_edge) ) {
     struct tcphdr* tcp_hdr = tcp_seg_tcphdr(seg);
 
+    /* This segment is a retransmission if:
+     * - its sequence number is not the next sequence number (i.e., we have
+     *   sent this sequence before, but have not yet had an ACK); or
+     * - the segment is marked as `in_flight`. This means that we have sent the
+     *   segment before, but have not yet processed a TX completion event for
+     *   the segment.
+     *
+     * Notably, we must store the state of `seg->in_flight` **BEFORE** we call
+     * `tcp_output_segment`, as that will result in `seg->in_flight` being set
+     * for all (successfully sent) segments.
+     */
+    const bool is_retransmit = (tcp_seg_seq(seg) != pcb->snd_nxt ||
+                                seg->in_flight);
+
     zf_assert_nequal(tcp_hdr->rst, 1);
 
     zf_log_tcp_tx_trace(tcp, "%s: snd_right_edge %u, cwnd %u, "
@@ -956,7 +970,7 @@ int tcp_output(struct zf_tcp* tcp)
       else if( tcp_hdr->fin ) {
         flags |= ZF_PKT_REPORT_TCP_FIN;
       }
-      if( tcp_seg_seq(seg) != pcb->snd_nxt ) {
+      if( is_retransmit ) {
         flags |= ZF_PKT_REPORT_TCP_RETRANS;
       }
       zf_tx_reports::prepare(&stack->tx_reports,
@@ -1032,14 +1046,13 @@ static int tcp_output_segment(struct zf_tcp* tcp, struct tcp_seg *seg)
   /* Stop zwin timer */
   zf_tcp_timers_timer_stop(tcp, ZF_TCP_TIMER_ZWIN);
 
-  /* Do not retransmit a segment which has not got out of NIC. */
+  /* If we have not yet processed a TX completion for this segment, then the
+   * packet might still exist in a buffer, so we should send it either by PIO
+   * or by copying the packet to a new buffer. */
   if( seg->in_flight ) {
     zf_log_tcp_tx_trace(tcp, "%s: in-flight %u:%u\n", __func__,
                         tcp_seg_seq(seg), tcp_seg_seq(seg) + seg->len);
-    /* Make sure the segment is sent in fire and forget mode,
-     * that is either with PIO or with a buffer copy to be freed at
-     * tx complete */
-    send_flags = ZF_REQ_ID_PIO_FLAG | ZF_REQ_ID_PROTO_TCP_FREE | PKT_INVALID;
+    send_flags = ZF_REQ_ID_PIO_FLAG | ZF_REQ_ID_PROTO_TCP_KEEP | PKT_INVALID;
   }
 
   tcp_output_timers_common(stack, tcp, tcp_seg_seq(seg));


ligallag-amd (approved):

src/lib/zf/tcp_out.c
```diff
@@ -956,7 +957,7 @@ int tcp_output(struct zf_tcp* tcp)
       else if( tcp_hdr->fin ) {
         flags |= ZF_PKT_REPORT_TCP_FIN;
       }
-      if( tcp_seg_seq(seg) != pcb->snd_nxt ) {
+      if( is_retransmit ) {
```
krishd-amd, line 960: This change is a bit non-obvious. Why did this have to be moved like this?

src/lib/zf/tcp_out.c
```diff
@@ -1032,14 +1033,13 @@ static int tcp_output_segment(struct zf_tcp* tcp, struct tcp_seg *seg)
   /* Stop zwin timer */
   zf_tcp_timers_timer_stop(tcp, ZF_TCP_TIMER_ZWIN);
 
-  /* Do not retransmit a segment which has not got out of NIC. */
```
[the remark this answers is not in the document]
ibeecraft-amd: This seems like an oversight in previous commits. The comment has not been valid for quite a while.
Originally no packet was sent in this case so comment was valid. Then change was made to actually send the packet in a 'fire and forget mode'. 

When TX timestamping was added this was potentially overlooked. Worth running the test with this patch in debug with logging enabled to ensure that the pkt buffers are still being free'd and re-used as expected.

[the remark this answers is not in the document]
jfeather-amd: This is indeed the case, thanks @ibeecraft-amd! I removed this comment because it misled me when I first read it. The new comment should better reflect our actual intention with this bit of code, and brings in the removed comment below that (as we wouldn't want to "forget" about this packet).

src/lib/zf/tcp_out.c
jfeather-amd: This specific check to know if we're retransmitting didn't actually need to be moved, but the important part is `seg->in_flight`. I felt this was a better way to represent the code rather than storing another bool, e.g., `seg_was_in_flight`. Importantly, we must capture the state of `seg->in_flight` __before__ the call to `tcp_output_segment` in this function, which will result in `seg->in_flight` being set. I will add a comment to the definition of `is_retransmit` to explain exactly what's going on so will defer explanation to that comment rather than this PR to make sure the code is understandable in absence of this PR.

src/lib/zf/tcp_out.c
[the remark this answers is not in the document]
jfeather-amd: For completeness, the above comment was:
- added in Xilinx-CNS/tcpdirect_history@da6bdc033fd97ad7cb7e769a427a0e178d534ec6
- invalid as of Xilinx-CNS/tcpdirect_history@2da2de10474981438afa593a9de826c9a40b4a50

And the failure mode observed here has been present since Xilinx-CNS/tcpdirect_history@59d0a4ec80c3d95b756d293a149c7a08e9e4c0ed.

src/lib/zf/tcp_out.c
```diff
@@ -1032,14 +1046,13 @@ static int tcp_output_segment(struct zf_tcp* tcp, struct tcp_seg *seg)
...
+   * packet might still exist in a buffer, so we should send it either by PIO
+   * or by copying the packet to a new buffer. */
   if( seg->in_flight ) {
     zf_log_tcp_tx_trace(tcp, "%s: in-flight %u:%u\n", __func__,
                         tcp_seg_seq(seg), tcp_seg_seq(seg) + seg->len);
-    /* Make sure the segment is sent in fire and forget mode,
-     * that is either with PIO or with a buffer copy to be freed at
-     * tx complete */
-    send_flags = ZF_REQ_ID_PIO_FLAG | ZF_REQ_ID_PROTO_TCP_FREE | PKT_INVALID;
+    send_flags = ZF_REQ_ID_PIO_FLAG | ZF_REQ_ID_PROTO_TCP_KEEP | PKT_INVALID;
```
jfeather-amd, line 1055: The change here is the "important" one, to use `ZF_REQ_ID_PROTO_TCP_KEEP` instead of `ZF_REQ_ID_PROTO_TCP_FREE`. That way, [`zf_reactor_handle_tx_timestamp_event`](https://github.com/Xilinx-CNS/tcpdirect/blob/ba607b3ca2ada1e56f96042fdf8552012aa785c7/src/lib/zf/private/reactor.c#L82) will call `zf_tx_reports::complete`.

src/lib/zf/tcp_out.c
```diff
@@ -956,7 +970,7 @@ int tcp_output(struct zf_tcp* tcp)
       else if( tcp_hdr->fin ) {
         flags |= ZF_PKT_REPORT_TCP_FIN;
       }
-      if( tcp_seg_seq(seg) != pcb->snd_nxt ) {
+      if( is_retransmit ) {
```
jfeather-amd, line 973: This updated logic to check if a packet is a retransmission is purely so that users will see the "this is a retransmission" flag in the TX report. For context, please see the following excerpt from the passing log:
```
Sending Payload: 65
#### LONG SLEEP HAPPENS HERE ####
Count of timestamps:1
Timestamps: start:64, bytes:65, flags:1, seconds:1721381680 nsecs:401902045
Count of timestamps:1
#### NOTE flags & 0x2000 DENOTES RETRANSMIT ####
Timestamps: start:64, bytes:65, flags:2001, seconds:1721381700 nsecs:402331909
```

krishd-amd (approved):

ibeecraft-amd (approved):
Ran some tests in the background and it does seem like this functions now as expected on my test-server.

Also just checked that the packets were being freed back to the pool correctly and it does seem so (confirmed with look at code path).

jfeather-amd:
Looks like I accidentally merged to master instead of v8_1, thanks @sianj-xilinx for pointing this out! I'll cherry-pick this to v8_1 then merge that branch to master.

PR #29, opened 2024-07-24
ON-15664: Generate SRPM with sources and not pre-built binaries

ivatet-amd:
Put TCPDirect source files instead of pre-built binaries into the SRPM package generated by the `zf_make_official_srpm` script.

It is quite a crude PR because, for instance, it does not list the tarball files explicitly and can catch temporary text editor files into the tarball, but there are more series to come to address it.

TODO:
- [x] Fix CI (test.groovy): Pass (Amber, in a personal pipeline)

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -20,6 +20,24 @@ Execute the following command from the root of the repository:
 export ONLOAD_TREE={path to the checked out onload repository}
 make
 ```
+
+### Packaging
+Use the [`zf_make_official_srpm`](./scripts/zf_make_official_srpm) script to create an SRPM package:
+
+```bash
+scripts/zf_make_official_srpm --version ${version}
+```
+
+The `${version}` variable can be git hexsha such as `d42f94d5` or a semver string such as `9.0.0`.
+
+Execute the following command to create an RPM package from SRPM:
+
+```bash
+rpmbuild --define "onload_tarball ${onload}" --rebuild ~/rpmbuild/SRPMS/tcpdirect-${version}-1.src.rpm
+```
+
+Here, the `${onload}` variable points to an Onload tarball, and `${version}` is the same as for `zf_make_official_srpm`.
+
 ## How to run unit tests
 
 ### Dependencies
--- a/scripts/tcpdirect_misc/tcpdirect.spec
+++ b/scripts/tcpdirect_misc/tcpdirect.spec
@@ -1,9 +1,16 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: (c) 2016-2023 Advanced Micro Devices, Inc.
 ######################################################################
+# RPM spec file for TCPDirect
+#
+# To build a source RPM:
+#   scripts/zf_make_official_srpm --version <version>
+#
+# To build a binary RPM from a source RPM:
+#   rpmbuild --define "onload_tarball <onload_tarball>" --rebuild ~/rpmbuild/SRPMS/tcpdirect-<version>-1.src.rpm
 
 %define _unpackaged_files_terminate_build 0
-pkgversion-DEFINITION
+%{!?pkgversion: %global pkgversion 9.0.0}
 
 Name:           tcpdirect
 Version:        %{pkgversion}
@@ -13,6 +20,8 @@ Summary:        TCPDirect
 License:        MIT AND BSD-3-Clause AND LGPL-3.0
 URL:            https://github.com/Xilinx-CNS/tcpdirect
 Source0:        tcpdirect-%{pkgversion}.tgz
+BuildRequires:  gcc make
+BuildRoot:      %{_builddir}/%{name}-root
 
 Vendor:         Advanced Micro Devices, Inc.
 
@@ -30,11 +39,13 @@ This package comprises the user space components of tcpdirect.
 [ "$RPM_BUILD_ROOT" != / ] && rm -rf "$RPM_BUILD_ROOT"
 mkdir -p $RPM_BUILD_ROOT
 
-
-
+%setup -n %{name}-%{pkgversion}
 
 %build
-tar -xzvf %{_sourcedir}/tcpdirect-%{pkgversion}.tgz -C $RPM_BUILD_ROOT
+scripts/zf_mkdist --version "%{pkgversion}" --name tcpdirect "%{?onload_tarball:%onload_tarball}"
+
+tar -xzvf build/tcpdirect-"%{pkgversion}".tgz -C "${RPM_BUILD_ROOT}"
+
 ls -la $RPM_BUILD_ROOT
 ls -la $RPM_BUILD_ROOT/tcpdirect-%{pkgversion}/scripts
 # $RPM_BUILD_ROOT/tcpdirect-%{pkgversion}/scripts/zf_install
--- a/scripts/zf_make_official_srpm
+++ b/scripts/zf_make_official_srpm
@@ -9,89 +9,88 @@ log()  { err "$me: $*"; }
 fail() { log "$*"; exit 1; }
 try()  { "$@" || fail "FAILED: $*"; }
 
+my_dir=$(cd "$(dirname "$0")" && /bin/pwd)
+top_dir=$(dirname "${my_dir}")
 
-usage_msg() {
-  err
-  err "usage:"
-  err "  $me <tcpdirect-tarball>"
-  err
-  err "options:"
-  err
-}
-
+tarball_files_from="
+versions.yaml
+Makefile
+Makefile.onload
+Makefile-top.inc
+doc
+src
+scripts/debian-templ
+scripts/tcpdirect_misc/tcpdirect-extract-notes
+scripts/zf_debug
+scripts/zf_install
+scripts/zf_make_official_deb
+scripts/zf_make_official_srpm
+scripts/zf_make_tarball
+scripts/zf_mkdist
+scripts/zf_uninstall
+"
 
 usage() {
-  usage_msg
+  echo
+  echo "usage:"
+  echo "  $(basename "$0") [options]"
+  echo
+  echo "options:"
+  echo "  --version <version>"
+  echo
   exit 1
 }
 
-
-logtry() {
-  log "$*"
-  try "$@"
-}
-
-
 cleanup() {
-  [ -d "$tmpd" ] && rm -rf "$tmpd"
+  echo "Deleting ${tmpdir}"
+  rm -rf "${tmpdir}"
 }
-
 
 ######################################################################
 # main()
+
 version=
-tarball=
 
 while [ $# -gt 0 ]; do
-    case "$1" in
-        --version)  shift; version="$1";;
-        -*)  usage;;
-        *)  tarball="$1";;
-    esac
-    shift
+  case "$1" in
+    --version)  shift; version="$1";;
+    *)  usage;;
+  esac
+  shift
 done
 
-if [ -z "$version" ]; then
+if [ -z "${version}" ]; then
     version=$(git rev-parse HEAD)
 fi
 
-# [ $# = 1 ] || usage
-# tarball="$1"
+tmpdir=$(mktemp -d "/tmp/${me}.XXXXXXXX")
+zf_prefix="tcpdirect-${version}"
+zf_spec="scripts/tcpdirect_misc/tcpdirect.spec"
+zf_tarball="${tmpdir}/${zf_prefix}.tar"
+zf_tgz="${tmpdir}/${zf_prefix}.tgz"
 
-tmpd=
 trap cleanup EXIT
-tmpd=$(mktemp -d /tmp/$me.XXXXXXXX)
 
-try [ -f "$tarball" ]
-tarballdir=$(cd "$(dirname "$tarball")" && /bin/pwd)
-tarballbase=$(basename "$tarball")
-tarball="$tarballdir/$tarballbase"
-try [ -f "$tarball" ]
+# Create a tarball in tmpdir.
+# shellcheck disable=SC2086
+try tar -c -C "${top_dir}" --transform "s,^,${zf_prefix}/,S" -f "${zf_tarball}" -- ${tarball_files_from}
 
-original_wd=$(/bin/pwd)
-try cd "$tmpd"
+# Patch and put the SPEC file in tmpdir.
+# shellcheck disable=SC2046
+mkdir -pv $(dirname "${tmpdir}/${zf_prefix}/${zf_spec}")
 
-# Edit the "Packager" line in the rpm spec file and rebuild tarball.
-try tar xf "$tarball"
-distd="$(basename "${tarball%.tgz}")"
-try [ -d "$distd" ]
-tab=$(echo -n -e "\t")
-try rm "$distd/scripts/zf_install" "$distd/scripts/zf_uninstall"
+try sed \
+  -E "s/^.*(%global pkgversion).*$/%define pkgversion ${version}/" "${top_dir}/${zf_spec}" \
+  > "${tmpdir}/${zf_prefix}/${zf_spec}"
 
-try sed -i \
- -e "s/pkgversion-DEFINITION/%define pkgversion $version/g" \
- "$distd/scripts/tcpdirect.spec"
+# Add the patched SPEC file in the tarball.
+try tar -v -C "${tmpdir}" --append -f "${zf_tarball}" -- "${zf_prefix}/${zf_spec}"
 
-tarball=$(basename "$tarball")
-try tar czf "$tarball" --owner=root --group=root "$distd"
+# Compress the tarball.
+try gzip -v "${zf_tarball}"
 
-# Build source rpm.
-try mkdir -p rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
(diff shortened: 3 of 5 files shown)


krishd-amd:
Putting the bash scripts into shellcheck seems to show some linting problems. Would you mind addressing them?

ivatet-amd:
Thanks @krishd-amd. I've shellchecked the entire `zf_make_official_srpm` and snippets from other files. The warnings mainly were about double-quoting variables `${var}` -> `"${var}"`, which I definitely agree with. Please ping me if your shellcheck spots more problems!

scripts/zf_mkdist
```diff
@@ -174,6 +175,7 @@ while [ $# -gt 0 ]; do
         --unstripped) stripped=false;;
         --shim) shim=true;;
         --name) shift; name="$1";;
+        --build_opts) shift; build_opts="$1";;
```
jfeather-amd, line 178: I feel like these kinds of multi-word command line options typically get separated with a `-` rather than `_` (so `--build-opts`). Very unimportant suggestion in reality, and perhaps it's just a personal preference thing anyway!

scripts/zf_mkdist
```diff
@@ -84,13 +85,13 @@ build_and_copy_artifacts() {
...
-        make_args=
+        make_args="${build_opts}"
     elif [ "${build_type}" == "shim" ]; then
         echo "Building socket shim"
-        make_args="ZF_DEVEL=1 shim"
+        make_args="${build_opts} ZF_DEVEL=1 shim"
     else
         echo "Building ndebug binaries"
-        make_args=NDEBUG=1
+        make_args="${build_opts} NDEBUG=1"
```
jfeather-amd, line 94: I imagine you've been very particular about positioning the user provided `build_opts` before those we want for a given build type, such that we overwrite whatever is provided in `build_opts`. Do you know if it's the case that `make` takes the value it sees last, e.g., so that `make A=1 A=3` would mean `make` only thinks `A=3`?

As an extension, I wonder if we should try to improve the fault-tolerance of these build types to explicitly exclude some things. For example, it might be good to define `NDEBUG=0` in "debug" to make sure these build targets can't be corrupted by bad user input.

scripts/zf_make_official_srpm
```diff
@@ -10,88 +10,89 @@ fail() { log "$*"; exit 1; }
...
+make_tarball() {
+  tarball="${zf_package}.tgz"
+
+  try tar -czf "${tarball}" -C "${tmpdir}" "${zf_package}"
+  try md5sum "${tarball}" > "${tarball}.md5"
+}
+
+
+make_srpm() {
+  tarball="${zf_package}.tgz"
```
jfeather-amd, line 67: Given this is defined in both `make_tarball` and `make_srpm`, it could be a good idea to just lump this with the other global definitions (e.g., right after `zf_package`).

scripts/zf_make_official_srpm
```diff
@@ -10,88 +10,89 @@ fail() { log "$*"; exit 1; }
...
+copy_files_to_tmpdir() {
+  try mkdir -p "${zf_tmpdir}"
+  # shellcheck disable=SC2086
+  try tar -cz -C "${top_dir}" ${copy_to_tmpdir} | tar xz -C "${zf_tmpdir}"
+}
+
+
+patch_tmpdir_files() {
+  try sed -i \
+    -e "s/pkgversion-DEFINITION/%define pkgversion ${version}/g" \
```
jfeather-amd, line 53: This is probably overthinking it, but can `version` contain `/`? I imagine in practice it's either a full version like `8.1.3` or a commit hash.

jfeather-amd (approved):
Looks good to me! I've left some comments, but believe all of these can be reasonably ignored.

scripts/zf_mkdist
ivatet-amd: Thanks, I think a `-` (dash) is indeed more common than `_` (underscore), and I can never get it right the first time!

scripts/zf_mkdist
ivatet-amd: In your example, I think `A=3` is the "strongest" and even beats the environment variables, i.e. `A=2 make A=1 A=3` should still think `A=3`.

My motivation to expose this variable was to enable parallelism. I probably would have taken a different path, such as what `onload_build` does - it computes the `parallel` variable automatically, but I noticed an unused `build_opts` and gave it a go.

I'm not keen on adding a new `debug` target in this specific case because:

1. We would probably need to handle `ZF_DEVEL=1` for the `shim` target the same way, and there might be more configuration knobs added soon?
2. It would make it look less similar to the Onload build scripts, which allow for the export of variables in envs. I would keep the Onload and TCPDirect scripts somewhat similar!

I see the use of the `parallel` variable as a safer alternative. What do you reckon?

scripts/zf_make_official_srpm
ivatet-amd: Yep, that sounds good.

(FWIW, I'm not keen on using global variables this way, but it makes perfect sense for smaller scripts like this!)

scripts/zf_make_official_srpm
ivatet-amd: If `version` contains `/`, `sed` fails:
```console
$ scripts/zf_make_official_srpm --version foo/bar
sed: -e expression #1, char 48: unknown option to `s'
zf_make_official_srpm: FAILED: sed -i -e s/pkgversion-DEFINITION/%define pkgversion foo/bar/g /tmp/zf_make_official_srpm.omlWw3bC/tcpdirect-foo/bar/scripts/tcpdirect_misc/tcpdirect.spec
Deleting /tmp/zf_make_official_srpm.omlWw3bC
```

If I escape it with `\/`, then `tar` fails later:
```console
$ scripts/zf_make_official_srpm --version 'foo\/bar'
tar (child): tcpdirect-foo\\/bar.tgz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: tcpdirect-foo\\/bar.tgz: Cannot write: Broken pipe
tar: Child returned status 2
tar: Error is not recoverable: exiting now
zf_make_official_srpm: FAILED: tar -czf tcpdirect-foo\/bar.tgz -C /tmp/zf_make_official_srpm.mBhY21vN tcpdirect-foo\/bar
Deleting /tmp/zf_make_official_srpm.mBhY21vN
```

We can sanitise the version string and prevent these "SQL injection"-like problems, but I'm quite relaxed about it in this context! However, if we want to sanitise it, I can think of `egrep -i '^[a-z0-9\-\_\.]+$'` as a filter to enable case-insensitive alpha-numerical input along with a `-` (dash), `_` (underscore), and `.` (full stop). What do you think?

scripts/tcpdirect_misc/tcpdirect.spec
```diff
@@ -1,6 +1,13 @@
...
+# RPM spec file for TCPDirect
+#
+# To build a source RPM:
+#   scripts/zf_make_official_srpm --version <version>
+#
+# To build a binary RPM from a source RPM:
+#   rpmbuild --define "onload_tarball <onload_tarball>" --rebuild ~/rpmbuild/SRPMS/tcpdirect-<version>-1.src.rpm
 
 %define _unpackaged_files_terminate_build 0
 pkgversion-DEFINITION
```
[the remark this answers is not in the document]
ivatet-amd: Thanks, it sounds good. I will apply your suggestion in the next update of the PR.

(I looked at it and didn't like it for the same reason, but didn't fix it because it wasn't on my way!)

scripts/tcpdirect_misc/tcpdirect.spec
```diff
@@ -13,6 +20,8 @@ Summary:        TCPDirect
 License:        MIT AND BSD-3-Clause AND LGPL-3.0
 URL:            https://github.com/Xilinx-CNS/tcpdirect
 Source0:        tcpdirect-%{pkgversion}.tgz
+BuildRequires:  gcc make
+BuildRoot:      %{_builddir}/%{name}-root
```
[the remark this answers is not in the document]
ivatet-amd: Yes, I am not motivated to get rid of `RPM_BUILD_ROOT` in this PR exactly due to the hackery later on in the snippet. I've captured this suggestion in ON-15949.

scripts/tcpdirect_misc/tcpdirect.spec
```diff
@@ -30,11 +39,13 @@ This package comprises the user space components of tcpdirect.
 [ "$RPM_BUILD_ROOT" != / ] && rm -rf "$RPM_BUILD_ROOT"
 mkdir -p $RPM_BUILD_ROOT
 
+%setup -n %{name}-%{pkgversion}
 
+%build
+scripts/zf_mkdist --build_opts "-j$(nproc)" --version "%{pkgversion}" --name tcpdirect "%{?onload_tarball:%onload_tarball}"
```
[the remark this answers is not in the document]
ivatet-amd: That's neat, I didn't know about it.

scripts/tcpdirect_misc/tcpdirect.spec
```diff
@@ -30,11 +39,13 @@ This package comprises the user space components of tcpdirect.
...
 
+tar -xzvf build/tcpdirect-"%{pkgversion}".tgz -C "${RPM_BUILD_ROOT}"
```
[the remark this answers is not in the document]
ivatet-amd: Currently, we do something unusual!

A few lines above, I've added `scripts/zf_mkdist` that consumes the content of an archive unpackaged by `%setup`. However, again due to the hackery later on, this line unpackages another/new archive created by `scripts/zf_mkdist` to consume by the rest of the snippet.

I limited the scope of this PR to literally making the SRPM genuine SRPM (in contrast to a package with pre-built binaries) and captured further work in ON-15695.

scripts/zf_make_official_srpm
jfeather-amd: My main concern here was that it could silently fail, so glad to hear that `sed` will fail. I agree with @pcolledg-amd's comment, it's probably sensible to just pop something in the README and let the script fall over in such cases.

scripts/zf_mkdist
jfeather-amd: > I see the use of the `parallel` variable as a safer alternative

Sounds good to me! It does sound like a tricky problem to handle properly otherwise.

scripts/zf_make_official_srpm
ivatet-amd: > CI/CD flows are almost always trusted

It accurately describes why I am relaxed about it!

scripts/tcpdirect_misc/tcpdirect.spec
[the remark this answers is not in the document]
ivatet-amd: Josh has convinced me to replace `--build_opts` with the parallelism embedded in `zf_mkdist`!

ivatet-amd:
Thank you for the prompt reviews! I've either addressed the comments as agreed or captured follow-up suggestions as tasks in ON-15695.

Overall, the SPEC file still requires a lot of care, but its improvements depend on some heavy lifting, such as refactoring Makefile and scripts like `zf_mkdist`.

README.MD
```diff
@@ -20,6 +20,24 @@ Execute the following command from the root of the repository:
 export ONLOAD_TREE={path to the checked out onload repository}
 make
 ```
+
+### Packaging
+Use the `zf_make_official_srpm` script to create an SRPM package:
```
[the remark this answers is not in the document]
ivatet-amd: Yup!

README.MD
```diff
@@ -20,6 +20,24 @@ Execute the following command from the root of the repository:
...
+```bash
+$ scripts/zf_make_official_srpm --version ${version}
+```
+
+The `${version}` variable can be git hexsha such as `d42f94d5` or a semver string such as `9.0.0`.
+
+Execute the following command to create an RPM package from SRPM:
+
+```bash
+$ rpmbuild --define "onload_tarball ${onload}" --rebuild ~/rpmbuild/SRPMS/tcpdirect-${version}-1.src.rpm
```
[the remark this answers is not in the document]
ivatet-amd: Yup.

scripts/zf_mkdist
```diff
@@ -165,6 +166,7 @@ stripped=true
 shim=false
 name="zf"
 build_opts=""
+parallel=-j$(nproc 2> /dev/null || echo 4)
```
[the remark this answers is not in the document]
ivatet-amd: Yup. (I took it as-is from the Onload scripts, so we might want to fix it over there, too.)

scripts/zf_make_official_srpm
```diff
@@ -10,88 +10,86 @@ fail() { log "$*"; exit 1; }
...
 }
 
 
-cleanup() {
-  [ -d "$tmpd" ] && rm -rf "$tmpd"
+copy_files_to_tmpdir() {
+  try mkdir -p "${zf_tmpdir}"
+  # shellcheck disable=SC2086
+  try tar -cz -C "${top_dir}" ${copy_to_tmpdir} | tar xz -C "${zf_tmpdir}"
+}
```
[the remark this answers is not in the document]
ivatet-amd: Yup, am applying your suggestion in the next revision.

This approach to tar/untar is inherited from the same function in zf_mkdist, which we should address over there too. Let me capture it in ON-15946.

I've reworked the workflow. Now it creates a tarball without the SPEC file in a temporary directory, copies and patches the SPEC file in the same temporary directory, appends it to the tarball and finally invokes `gzip` for compression. The `gzip` neatly renames the file, but I still want to `mv` to replace the file extension.

It also addresses the second/extraction `tar` invocation that didn't have `try` before it.

scripts/zf_make_official_srpm
```diff
@@ -10,88 +10,86 @@ fail() { log "$*"; exit 1; }
...
-  err "  $me <tcpdirect-tarball>"
-  err
-  err "options:"
-  err
-}
-
+my_dir=$(cd "$(dirname "$0")" && /bin/pwd)
+top_dir=$(dirname "${my_dir}")
+
+copy_to_tmpdir="
```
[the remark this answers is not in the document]
ivatet-amd: NP!

ivatet-amd:
Thanks for the review @pcolledg-amd. The `git-archive` approach is certainly an improvement over `tar`, but I'm not applying it here because it wouldn't work with the CI as-is: it doesn't stash the `.git` subdirectory.

The latest PR addresses your inline comments.

There is one pending problem with the binary RPM workflow in this PR, which I will fix shortly. (It won't substantially change the PR.)

tcrawley-xilinx (approved):

scripts/zf_make_official_srpm
```diff
@@ -9,89 +9,88 @@ log()  { err "$me: $*"; }
...
-srpmbase=$(basename "$srpm")
-destfile="$original_wd/$(echo "$srpmbase" | sed -e 's/\.el6//')"
-[ -f "$destfile" ] &&
-  fail "ERROR: '$destfile' already exists.  I will not overwrite."
-try cp "$srpm" "$destfile"
-echo "$destfile"
+
+# Create a tarball in tmpdir.
+# shellcheck disable=SC2086
+try tar -c -C "${top_dir}" --transform "s,^,${zf_prefix}/," -f "${zf_tarball}" -- ${tarball_files_from}
```
krishd-amd, line 76: Do these need to have shellcheck disabled?

ivatet-amd: `shellcheck` suggests double quotes around `${tarball_files_from}` we don't want them:

```console
$ foo="
> Makefile
> Makefile.onload
> "

$ ls "${foo}"
ls: cannot access ''$'\n''Makefile'$'\n''Makefile.onload'$'\n': No such file or directory

$ ls ${foo}
Makefile  Makefile.onload
```

I believe that with double quotes, the variable is interpreted as one parameter/file.

One alternative would be using `-T`/`--files-from` with the list in a file, or, even better, `git-archive`, but it seems to require patching the CI pipelines. I've put a comment with a suggestion about `git-archive` and test.groovy in ON-15946.

ivatet-amd:
The latest push addresses the tar/symlink transformation issue and adds the `S` suffix to the expression that prepends the `tcpdirect-${version}` prefix to all files in the tarball. Without it, transformation applies to the symlink _destination_ and messes it up.

Before (a broken symlink):
```console
$ ls -l tcpdirect-foo/src/tests/zf_apps/ | grep Makefile-top
lrwxrwxrwx 1 iteterev sflr    39 Mar 27 10:25 Makefile-top.inc -> tcpdirect-foo/../../../Makefile-top.inc
```

After (👍):
```console
$ ls -l tcpdirect-foo/src/tests/zf_apps/ | grep Makefile-top
lrwxrwxrwx 1 iteterev sflr    25 Mar 27 10:25 Makefile-top.inc -> ../../../Makefile-top.inc
```

https://www.gnu.org/software/tar/manual/html_node/transform.html

The latest personal pipeline is amber/green, so planning to merge it shortly. Thanks for the reviews!

scripts/tcpdirect_misc/tcpdirect.spec
abower-amd, line 24: Why did you add `BuildRoot`? This is forbidden in Fedora since EL6.

ivatet-amd: It must be a remnant of some previous work. Fixed in https://github.com/Xilinx-CNS/tcpdirect/pull/31.

PR #22, opened 2024-06-19
ON-12878: use the new cplane user API

jfeather-amd:
Starting with Xilinx-CNS/onload@b660ac5b259f33cddcb1c6bd6f81ef7910dc10a7, @rhughes-xilinx added a public cplane API to onload. This PR modifies tcpdirect to use this new public cplane API, and is ported from before the open-sourcing of tcpdirect. For internal reviewers, the following PRs may be useful reading as an introduction to this PR:
- Xilinx-CNS/tcpdirect_history#226, the original version of this PR in the archived closed-source tcpdirect.
- Xilinx-CNS/onload_internal#1332, the addition of the public cplane API to onload.

Many of the changes from the original PR are rebase conflicts, and some small bugs (mostly in the unit tests).

### Testing Done
- [x] Unit tests are happy (locally and on GitHub actions)
- [x] Send to multicast address
    - [x] Normal (`ZF_ATTR="interface=enp4s0f0" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.129.192:8080 239.1.2.3:8080`)
    - [x] VLAN (`ZF_ATTR="interface=enp4s0f0.1" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.111.1:8080 239.1.2.3:8080`)
- [x] Send to unicast address
    - [x] Normal (`ZF_ATTR="interface=enp4s0f0" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.129.192:8080 192.168.129.191:8080`)
    - [x] VLAN (`ZF_ATTR="interface=enp4s0f0.1" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.111.1:8080 192.168.111.2:8080`)
- [x] Send over bond
    - [x] LACP
        - [x] Unicast to other bond (`ZF_ATTR="interface=bond0" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.192.1:8080 192.168.192.2:8080`)
        - [x] Multicast (`ZF_ATTR="interface=bond0" ./build/gnu_x86_64/bin/zf_apps/static/zfsend -i 1 192.168.192.1:8080 239.1.2.3:8080`)
        - [x] Handle bonding changes
[...]

PR #30, opened 2024-07-26
ON-15945: Build tcpdirect against system installed onload

tcrawley-xilinx:
~~Built ontop of PR #22, but with those changes rebased onto the newest version of master. When #22 is rebased/merged I'll remove these extra commits, for now just focus on the 2 most recent commits.~~

### [1/2] Cleanup: Allow for relative path in ONLOAD_TREE

Not technically needed, but now something like `ONLOAD_TREE=../onload make` should work.

### [2/2] ON-15945: Build tcpdirect against system installed onload
If ONLOAD_TREE is set then the build will proceed as before, otherwise
now it will try to build against system installed libraries/headers.

Building with ZF_DEVEL=1 will still require setting ONLOAD_TREE due
since it includes some files from within onload's src tree.

## Testing done
Installed `onload` and `onload-devel` rpms, then built tcpdirect with `make`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,16 @@
 .DEFAULT_GOAL := all
 
 TOP := $(patsubst %/,%,$(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
-
+ifdef ONLOAD_TREE
 include Makefile.onload
+else
+  ifeq ($(ZF_DEVEL),1)
+    $(error ZF_DEVEL unsupported when building from installed onload)
+  endif
+  $(info Using installed onload libraries and headers)
+  CITOOLS_LIB := /usr/lib64/libcitools1.a
+  CIUL_LIB    := /usr/lib64/libciul1.a
+endif
 include Makefile-top.inc
 
 # Use Onload's build-tree structure, unless overridden.
--- a/Makefile.onload
+++ b/Makefile.onload
@@ -9,7 +9,7 @@ endif
 
 .PHONY: build_dir
 
-ONLOAD_SCRIPTS := $(ONLOAD_TREE)/scripts
+ONLOAD_SCRIPTS := $(shell realpath $(ONLOAD_TREE))/scripts
 export PATH := $(ONLOAD_SCRIPTS):$(PATH)
 
 
--- a/src/lib/zf/Makefile.inc
+++ b/src/lib/zf/Makefile.inc
@@ -50,23 +50,25 @@ ZF_GIT_BRANCH_INFO  := $(shell git symbolic-ref -q --short HEAD)
 public_objects: $(PUBLIC_OBJS)
 .PHONY: public_objects
 
+ifdef ONLOAD_TREE
+  ONLOAD_CFLAGS  := -I$(ONLOAD_TREE)/src/include \
+                    -I$(ONLOAD_UL_BUILD_DIR)/include \
+                    '-DONLOAD_VERSION_HDR="onload_version.h"'
+else
+  ONLOAD_CFLAGS  := '-DONLOAD_VERSION_HDR=<onload/onload_version.h>'
+endif
 # ZF sources are C files that are built as C++ using the cc command.  The
 # reasons being:
 #  * files can stay .c
 #  * gcc does not enable exceptions nor link with stdcpp library (as opposed to
 #    C++).
-# XXX: ZF has no business requiring TRANSPORT_CONFIG_OPT_HDR.  We're pulling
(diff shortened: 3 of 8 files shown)


ivatet-amd (approved):
The last two patches in the series look good in principle.

Curiously, that these files don't seem to require patching:
```console
$ git grep ONLOAD_TREE
...
src/tests/trade_sim/Makefile.inc:ifdef ONLOAD_TREE
src/tests/trade_sim/Makefile.inc:  ONLOAD_CFLAGS := -I$(ONLOAD_TREE)/src/include
src/tests/zf_apps/Makefile.inc:ifdef ONLOAD_TREE
src/tests/zf_apps/Makefile.inc:  ONLOAD_CFLAGS := -I$(ONLOAD_TREE)/src/include
src/tests/zf_internal/Makefile.inc:ifdef ONLOAD_TREE
src/tests/zf_internal/Makefile.inc:  ONLOAD_CFLAGS := -I$(ONLOAD_TREE)/src/include
```

Presumably, that's because they are included in the top-level Makefile at the point where flags are defined.

When I try and build them standalone (in `master`), the build fails:
```console
$ make -C src/tests/zf_apps
src/tests/zf_apps/zfsink.c:14:10: fatal error: zf/zf.h: No such file or directory
```

It makes me think that test apps can be built standalone only when TCPDirect and Onload are installed system-wide.

Would you accept a suggestion to remove the `ONLOAD_TREE` handling from these testing Makefiles as they seem to achieve nothing, and its handling is in the spirit of this PR?

Makefile.onload
```diff
@@ -9,7 +9,7 @@ endif
 
 .PHONY: build_dir
 
-ONLOAD_SCRIPTS := $(ONLOAD_TREE)/scripts
+ONLOAD_SCRIPTS := $(shell cd $(ONLOAD_TREE) && /bin/pwd)/scripts
```
jfeather-amd, line 12: Very much a nit (and personal preference), but this might be cleaner to represent as `realpath $(ONLOAD_TREE)` instead of the `cd` and `pwd` combo.

Makefile
```diff
@@ -6,8 +6,16 @@
...
-
+ifdef ONLOAD_TREE
 include Makefile.onload
+else
+ifeq ($(ZF_DEVEL),1)
+  $(error ZF_DEVEL unsupported when building from installed onload)
+endif
+$(info Using installed onload libraries and headers)
+CITOOLS_LIB := /usr/lib64/libcitools1.a
+CIUL_LIB    := /usr/lib64/libciul1.a
```
jfeather-amd, line 17: Could be nice to indent these bits (with spaces as to not upset make) if possible.

jfeather-amd (approved):
Looks good!

> Building with ZF_DEVEL=1 will still require setting ONLOAD_TREE due
since it includes some files from within onload's src tree.

Is the main blocker on this zf_emu? Is this is fully blocked in the sense that the files we need aren't installed either by onload?

tcrawley-xilinx:
> > Building with ZF_DEVEL=1 will still require setting ONLOAD_TREE due
> > since it includes some files from within onload's src tree.
> 
> Is the main blocker on this zf_emu? Is this is fully blocked in the sense that the files we need aren't installed either by onload?

The main issue I found was with the superbuf stuff for zf emu which pulls in a header file from outside the include tree.
e.g. https://github.com/Xilinx-CNS/tcpdirect/blob/master/src/lib/zf/private/zf_emu_superbuf.h#L87

There may be other examples, but this was just the first I encountered.

jfeather-amd:
> pulls in a header file from outside the include tree

I wonder if it's worth a follow up to move any such cases to the `include` tree, although perhaps out-of-scope for this PR.

src/lib/zf/Makefile.inc
```diff
@@ -50,6 +50,13 @@ ZF_GIT_BRANCH_INFO  := $(shell git symbolic-ref -q --short HEAD)
 public_objects: $(PUBLIC_OBJS)
 .PHONY: public_objects
 
+ifdef ONLOAD_TREE
+  ONLOAD_CFLAGS  := -I$(ONLOAD_TREE)/src/include \
+                    -I$(ONLOAD_UL_BUILD_DIR)/include \
```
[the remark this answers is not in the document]
tcrawley-xilinx: I don't think the definition is necessary any more. I've removed both instances of it (and the corresponding comments) and have not encountered any build issues (including building with ZF_DEVEL)

PR #32, opened 2024-08-05
ON-15945: Add option to build RPM from system installed onload

tcrawley-xilinx:
Previously, in order to build the tcpdirect binary rpm you would have to define the `onload_tarball` macro so that the build system would have all the requisite libraries and headers. Now that we have the `onload-devel` package (and have included citools.a in `onload`), we can use the system installed onload when building tcpdirect and do away with the macro.

The functionality of `onload_tarball` is still preserved in case a user wants to build from a specific version of onload but doesn't want to install it on their system. Additionally, if you want to build the socket shim using `zf_mkdist` you must specify the onload tarball to use as the shim requies `ZF_DEVEL=1` which in turn requires a full copy of the onload source code.

## Testing done
If `onload` and `onload-devel` packages were installed on the system I could successfully run `zf_mkdist` without specifying an onload tarball, and `rpmbuild --rebuild tcpdirect-*.src.rpm` was sufficient to build the binary rpm.
Otherwise the build system works the same as before.

### DIFF
--- a/scripts/zf_mkdist
+++ b/scripts/zf_mkdist
@@ -28,7 +28,9 @@ scripts
 usage() {
     echo
     echo "usage:"
-    echo "  $(basename $0) [options] <onload_tarball>"
+    echo "  $(basename $0) [options] [onload_tarball]"
+    echo
+    echo "(Installed onload & onload-devel utilised when onload_tarball omitted.)"
     echo
     echo "options:"
     echo "  --version <version>"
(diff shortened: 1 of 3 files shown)


scripts/zf_mkdist
```diff
@@ -186,7 +196,12 @@ if [ -z "$version" ]; then
     version=$(git rev-parse HEAD)
 fi
 
-[ $tarball_count -eq 1 ] || usage
+[ $tarball_count -eq 1 ] || [ -d /usr/include/ci ] || usage
+
+if $shim && [ -z "$onload_tarball" ]; then
```
jfeather-amd, line 201: Not sure if this file is already shellchecked, but is shellcheck fine with `$shim` by itself?

jfeather-amd (approved):
Looks good to me!

scripts/zf_mkdist
tcrawley-xilinx: Doesn't complain on this line. Raised #33 for the existing issues

PR #35, opened 2024-08-20
ON-15999: Fix build failure caused by ON-15867

dchadwic-xilinx:
Build error resolved by patch


jfeather-amd (approved):
Thanks! Manually tested this as GHA wasn't using the latest onload 8 branch. I'd like to wait for my PR #36 to be merged before merging this so that this doesn't get falsely reported as a failing commit.

PR #36, opened 2024-08-20
ON-15724: replace release tag with alternative version

jfeather-amd:
Now that Xilinx-CNS/onload@v8_1 exists alongside the branch Xilinx-CNS/tcpdirect@v8_1, we want to use that for builds and testing instead. This change also renames the release tag mechanism with an alternative version string as that makes more sense.

### Testing Done
GHA [now fails](https://github.com/jfeather-amd/tcpdirect/actions/runs/10470331764/job/28995268477) (as expected), fixed by #35.

### DIFF
--- a/.github/workflows/perform_build.yml
+++ b/.github/workflows/perform_build.yml
@@ -14,8 +14,7 @@ jobs:
     timeout-minutes: 1
     outputs:
       tcpdirectBranch: ${{ steps.tcpdirectBranch.outputs.branch }}
-      onloadBranch: ${{ steps.onloadBranch.outputs.branch }}
-      onloadReleaseTag: ${{ steps.onloadReleaseTag.outputs.tag }}
+      onloadPublicTreeish: ${{ steps.onloadPublicTreeish.outputs.tag }}
       packetDrillBranch: ${{ steps.packetDrillBranch.outputs.branch }}
     steps:
       - name: tcpdirect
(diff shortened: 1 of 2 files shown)


ligallag-amd (approved):
Changes seem fine. Not sure why you've chosen to name it "alternative". As I understand it, v8_1 is the release tag. That said, I frequently find myself being wrong in these areas!!!

sianj-xilinx:
> Changes seem fine. Not sure why you've chosen to name it "alternative". As I understand it, v8_1 is the release tag. That said, I frequently find myself being wrong in these areas!!!

I looked at this, was confused by what "alternative" was supposed to be, but didn't feel like I knew enough about the GHA stuff to complain!

ivatet-amd:
Would `branch_name` or `tree_ish` be more specific?

jfeather-amd:
Thanks for the review!

> Not sure why you've chosen to name it "alternative". As I understand it, v8_1 is the release tag.

I think "alternative" is quite a lazy name here honestly, but I wanted it to be sufficiently general as it could reasonably be a tag or a branch. Previously, the public tags (i.e., OpenOnload release builds) were used, whereas now we are using the v8_1 development branch (which corresponds to the internal onload-8.1 branch). The need for having an "alternative" is that we need to keep the reference to "onload-8.1" for Jenkins, but want to use "v8_1" in GHA. In this case, I felt it would just be better to try both and avoid potential complications when merging to master.

jfeather-amd:
> Would branch_name or tree_ish be more specific?

I think `tree_ish` is probably the most appropriate here, thanks!

jfeather-amd:
Renamed to `public_tree_ish`. Also opted to drop this being a "backup" mechanism as it was before, we should just maintain this in `versions.yaml` to avoid the extra overhead per GHA run.

Still [failing as expected](https://github.com/jfeather-amd/tcpdirect/actions/runs/10487076972/job/29046692513).

PR #41, opened 2024-09-30
ON-16081: update installation instructions in release notes

jfeather-amd:
The existing instruction were sufficient for those cases, and the only new one is DEB packaging AFAIK.

### DIFF
--- a/doc/ReleaseNotes
+++ b/doc/ReleaseNotes
@@ -30,7 +30,7 @@ TCPDirect distributed as source code
   package.
 
 
-Installation (TBD: update this section)
+Installation
 ------------
 
   To install TCPDirect from a tarball:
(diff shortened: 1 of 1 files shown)


tcrawley-xilinx (commented):
Could you make a note of the dependencies (e.g. `openonload` and `openonload-devel` needed by tcpdirect srpm)

jfeather-amd:
Latest push documents the approach-specific package dependencies.

tcrawley-xilinx (approved):

abower-amd:
> Taking the latest artefact from Jenkins, that doesn't appear to be the case:

I see a new Jira issue appearing in my crystal ball!

abower-amd (approved):
Post-commit further revision will be required when we have made sure the three release packages are all source releases!

krishd-amd:
I suggest for a future revision we perhaps drop the talk about onload-devel and onload-user packages. The install process does rather scream about the dependencies needed anyway.

PR #43, opened 2024-10-01
ON-15664: Produce debian source package

tcrawley-xilinx:
This PR adds a new script `zf_mksrc` which creates a tarball that is then used to create the deb package. This tarball can also be archived in the CI pipeline as a record of the source code that isn't either an rpm or deb (though there are current discussions as to whether this is desirable or not).

`onload_make_official_deb` has been updated to take a tarball created by `zf_mksrc` and produce a debiansource tarball/package from this.

The deb creation process is quite similar to the onload script.

I had a quick look at changing `debian-templ` -> `debian`, but I couldn't think of a nice way to avoid using `#VERSION#` in the rules file.

There are some potential improvements such as using `dch` in `zf_make_official_deb` to generate a changelog with an appropriate version number at package creation time. I want to think build solution that doesn't use `zf_mkdist`, but I don't think that blocks these changes so I'll raise it as a future PR.

See Xilinx-CNS/onload_internal#1702 for details of `onload-dev` package.

TODO:
- [x] Update build and install dependencies for new packages

---

### [3/3] ON-16081: Update ReleaseNotes and README with new deb instructions

---

## Testing done

Manually created and installed packages on ubu24.04

TODO:
- [x] Verify SRPM installation unaffected
- [x] Jenkins pipeline
- [x] Github actions


scripts/debian-templ/rules
```diff
@@ -0,0 +1,44 @@
...
+export DEB_BUILD_MAINT_OPTIONS=optimize=-lto
+
+# See FEATURE AREAS in dpkg-buildflags(1).
+#export DEB_BUILD_MAINT_OPTIONS = hardening=+all
+
+# See ENVIRONMENT in dpkg-buildflags(1).
+# Package maintainers to append CFLAGS.
+#export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
+# Package maintainers to append LDFLAGS.
+#export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
```
abower-amd, line 19: Have you checked that the net effect of using Debian's default option set is reasonable compared with what our own makefiles are doing and that our own Makefile's options are taking effect as intended? Our options have obviously been carefully chosen although if it's possible to mix in Debian's hardening and optimisations without loss of performance and functionality that is obviously ideal ultimately, if not now.

scripts/debian-templ/rules
```diff
@@ -0,0 +1,44 @@
...
+# includes the example source and header files, which prevents easy compilation.
+# Exclude the examples from the compress so that they are still compilable.
+override_dh_compress:
+	dh_compress -Xexamples
+
+# dh_make generated override targets.
+# This is an example for Cmake (see <https://bugs.debian.org/641051>).
+#override_dh_auto_configure:
+#	dh_auto_configure -- \
+#	-DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)
```
abower-amd, line 44: shall we just delete stuff we don't need?

scripts/zf_mksrc
```diff
@@ -0,0 +1,52 @@
...
+    *)  usage;;
+  esac
+  shift
+done
+
+if [ -z "${version}" ]; then
+    version=$(git rev-parse HEAD)
+fi
+
+tmpdir=$(mktemp -d "/tmp/${me}.XXXXXXXX")
```
abower-amd, line 44: Ideally we would stop creating temp directories like this and build under the current directory as normally expected in build operations (or one above in Debian's case!) but we can fix this at some future time.

abower-amd (approved):

scripts/debian-templ/rules
tcrawley-xilinx: I haven't checked anything, I suppose I could check what our zf performance tests do and test that manually. Do you know if there is a way to unset all debian build flags? Neither google nor `man dpkg-buildflags` gave a good option beyond disabling all areas individually

tcrawley-xilinx:
Didn't mean to include any changes to `ci/` yet. Please ignore for the time being

scripts/debian-templ/rules
abower-amd: I don't know but initially shall we just compare the command lines from tgz build and deb build and see how flags differ?

krishd-amd (approved):
Code seems fine, although i still am not sure of the motivation for the source package versus just producing debs.

scripts/debian-templ/rules
tcrawley-xilinx: When I was tracking down the LTO stuff, which resulted in the disabling on line 10, I compared the output from `make` for the problematic `cc` command, but there was no difference. I guess it was being picked up from the environment variables.

ci/test.groovy
```diff
@@ -181,6 +181,7 @@ nm.slack_notify() {
       """
       extractNotes("tcpdirect/scripts", "tcpdirect/build/tcpdirect-${tcpdirect_version_long}.tgz")
       dir('tcpdirect/build/') {
+        // TBD: Do we want these still?
```
abower-amd, line 184: Yes; they will be archived and available 'on demand'. (For now, at least.)

tcrawley-xilinx:
jenkins CI updated to now produce a debian source package.
Also updated build requirements so it needs `onload`, `onload-dev` and `build-essential`. For some reason we don't provide a nice virtual package from the onload deb, so it uses an OR of all the possible `onload` and `onload-dev` variations. It would be nice to improve this, but I don't think it is essential.

PR #46, opened 2024-10-03
Restore release notes artifact, while not including it in package zips

tcrawley-xilinx:
## Testing done
Personal jenkins pipeline

### DIFF
--- a/ci/test.groovy
+++ b/ci/test.groovy
@@ -192,6 +192,7 @@ nm.slack_notify() {
         archiveArtifacts(artifacts: '*.tgz')
         archiveArtifacts(artifacts: '*.md5')
         archiveArtifacts(artifacts: '*.txt')
+        sh 'rm *ReleaseNotes.txt'
         stash name: 'text_files', includes: '*.txt'
         zip_and_archive_files(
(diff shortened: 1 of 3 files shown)


krishd-amd (approved):

ci/test.groovy
```diff
@@ -192,7 +192,7 @@ nm.slack_notify() {
         archiveArtifacts(artifacts: '*.tgz')
         archiveArtifacts(artifacts: '*.md5')
         archiveArtifacts(artifacts: '*.txt')
-        stash name: 'text_files', includes: '*.txt'
+        stash name: 'text_files', includes: '*.txt', excludes: '*ReleaseNotes.txt'
```
jfeather-amd, line 195: Do we care that this will still result in it being included in the `*tarball-doxbox.zip`?

tcrawley-xilinx: 👍 Wasn't sure if we cared about the tarball or not. I'll update it

abower-amd: Oh I see, sorry I hadn't clocked that was the only one of the three affect. Probably worth getting right to retain the option.

jfeather-amd (approved):

PR #47, opened 2024-10-03
ON-160xx: Proposed release notes addition mentioning onload-dev

dchadwic-xilinx:
Any thoughts?

### DIFF
--- a/doc/ReleaseNotes
+++ b/doc/ReleaseNotes
@@ -31,6 +31,17 @@ TCPDirect distributed as source code
   packages are suitable for all supported OSs.
 
 
+New onload-devel package required
+---------------------------------
+
+  Onload now includes a package containing headers required by ef_vi
+  applications like TCPDirect. Installing this package is now required to
+  install TCPDirect binaries as well as to build TCPDirect applications.
+
+  Headers can also be installed via new '--headers' option in 'onload_install'
+  as of Onload-9.0.0.
+
+
 Installation
 ------------
 


doc/ReleaseNotes
```diff
@@ -31,6 +31,16 @@ TCPDirect distributed as source code
   packages are suitable for all supported OSs.
 
 
+New onload-devel package required
+---------------------------------
+
+  Onload now includes a package containing all of the header files used by
+  Onload. Installing this package is now required to build TCPDirect applications.
+
+  Headers can also be installed via new '--headers' option in 'onload_install'
```
matthewr-xilinx, line 40: Maybe omit this? TCPDirect-9 will be provided as SRPM/ source DEB only. So my current expectation is that we want the customer to use the same type of Onlaod-9 package?

doc/ReleaseNotes
```diff
@@ -31,6 +31,16 @@ TCPDirect distributed as source code
   packages are suitable for all supported OSs.
 
 
+New onload-devel package required
+---------------------------------
+
+  Onload now includes a package containing all of the header files used by
+  Onload. Installing this package is now required to build TCPDirect applications.
```
abower-amd, line 38: I'm ok with this. As a subsequent editorial pass I'd be tempted to change to "containing headers required by ef_vi applications like TCP Direct" or similiar, as it may not always be "all of the header files used by Onload" - it's not actually ideal that we include headers that are really internal and that could change.

matthewr-xilinx, line 38: Should this be `TCPDirect binaries` instead (or as well as) `TCPDirect applications` ?

matthewr-xilinx (changes requested):

doc/ReleaseNotes
sianj-xilinx: > I'd be tempted to change to "containing headers required by ef_vi applications like TCP Direct" or similiar, as it may not always be "all of the header files used by Onload" - it's not actually ideal that we include headers that are really internal and that could change.

I strongly agree with this. I hadn't been paying enough attention to the packaging changes and only recently noticed that it was installing all the headers. If I'd noticed earlier I'd have grumbled about it at the time, but then deemed it too late by the point I noticed. This is definitely something that should be fixed before the next release though.

doc/ReleaseNotes
dchadwic-xilinx: It can be downloaded from the Git checkout no? And a tarball generated from the checkout?

doc/ReleaseNotes
dchadwic-xilinx: Good point (as well as). But the reasoning for it is before the install steps so maybe need to mention the TCPDirect binaries on priority

doc/ReleaseNotes
dchadwic-xilinx: And its safe to assume customers will not always use the same package type

doc/ReleaseNotes
dchadwic-xilinx: Agreed. Thanks

abower-amd (changes requested):
I've changed my mind on supporting this patch - there's too much overlap with the first section on "Public Onload/ef_vi control plane API".

One resolution would be to combine the two sections into a more holistic explanation (a bit like in https://github.com/Xilinx-CNS/onload_internal/pull/1707).

Another approach would be that as building from source is completely new, the user will *have* to follow the installation instructions below which clearly already call out the need for the devel packages making this paragraph redundant!

dchadwic-xilinx:
@abower-amd 
> as building from source is completely new, the user will have to follow the installation instructions below which clearly already call out the need for the devel packages

My expectation is that customers who are used to the old way will now be wondering _where_ exactly they should be getting this onload-dev package from. They have not yet learned that this is provided using their current rpmbuild / debuild method, and as a result they may not notice the package sitting among their Onload package binaries (which is the mistake I made!)

dchadwic-xilinx:
@abower-amd 
>there's too much overlap with the first section on "Public Onload/ef_vi control plane API".

```
  The Onload control plane in this Onload-9.0.0 is presented via a new
  public API that can be used by ef_vi applications. As an ef_vi application,
  TCPDirect now uses this API to query the control plane server.
```
From my perspective as someone less knowledgeable about the implication of this statement: There is no overlap with the statements from the proposed "New onload-devel package required" section.

abower-amd:
> @abower-amd
> 
> > there's too much overlap with the first section on "Public Onload/ef_vi control plane API".
> 
> ```
>   The Onload control plane in this Onload-9.0.0 is presented via a new
>   public API that can be used by ef_vi applications. As an ef_vi application,
>   TCPDirect now uses this API to query the control plane server.
> ```
> 
> From my perspective as someone less knowledgeable about the implication of this statement: There is no overlap with the statements from the proposed "New onload-devel package required" section.

The overlap is in the implied dependence on exported headers, but really the underlying commonality is in the fact that TCPDirect is now distributed at source, which was one of the motivations for the independent control plane API, hence why a holistic treatment is more appropriate.

> @matthewr-xilinx any thoughts on whether or not a customer is likely to make the connection between this info and the need to install a new onload-dev package?

I didn't say the presence of this paragraph obviated the need to describe the requirement to install the devel package, although I did of course go on to suggest that need was covered in the installation steps!

matthewr-xilinx:
> @matthewr-xilinx any thoughts on whether or not a customer is likely to make the connection between this info and the need to install a new onload-dev package?

Need to be explicit that onload-dev/devel must be installed to build TCPDirect. Key thing is for this to be covered in the install section. But might be useful to add to the "Public Onload/ef_vi control plane API" section too, so there is more chance people will spot it. As you say, not all customers will have the insight to realise that this is required.

abower-amd:
Thanks for kicking off this revision, now incorporated in #48!

PR #48, opened 2024-10-03
cleanup: iterate on pr/47 onload-dev clarification

abower-amd:
Apologies for the second PR. I've taken @dchadwic-xilinx's suggestion, reworded commit log (no point in issue number that doesn't exist) then reordered the release notes and woven the onload-devel theme through.

### DIFF
--- a/doc/ReleaseNotes
+++ b/doc/ReleaseNotes
@@ -9,14 +9,6 @@ TCPDirect-9.0.0
   derives from those supported by Onload.
 
 
-Public Onload/ef_vi control plane API
--------------------------------------
-
-  The Onload control plane in this Onload-9.0.0 is presented via a new
-  public API that can be used by ef_vi applications. As an ef_vi application,
-  TCPDirect now uses this API to query the control plane server.
-
-
 TCPDirect distributed as source code
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


doc/ReleaseNotes
```diff
@@ -27,17 +19,40 @@ TCPDirect distributed as source code
...
+  Please ensure the openonload-devel RPM or onload-dev DEB is installed
+  after building Onload before attempting to build TCPDirect.
+
+
+Public Onload/ef_vi control plane API
+-------------------------------------
+
+  The Onload control plane in this Onload-9.0.0 is presented via a new
+  public API that can be used by ef_vi applications. As an ef_vi application,
+  TCPDirect now uses this API to query the control plane server.
```
dchadwic-xilinx, line 42: Can we swap the order of these two sections? If a customer gets stuck on the missing `onload-dev` message when installing, their first thought may be "_Did I miss a step?_" so would be best if its immediately before the installation section imo.

dchadwic-xilinx (changes requested):

doc/ReleaseNotes
abower-amd: I already switched it to this way round to have the new package section flow naturally from

> The packaging for Onload and TCPDirect has been refreshed to allow suitable new build and installation workflows for the TCPDirect source package."

Do you really think the idea hasn't been called out enough by the time they reach the installation instructions and the very verbose point 0?

dchadwic-xilinx: Its more because it answers the question of where a customer gets onload-dev from:
>  Please ensure the openonload-devel RPM or onload-dev DEB is installed
  **after building Onload** before attempting to build TCPDirect.

This sentence says you get it from building Onload (rather than some other source).

abower-amd: I think we are in the margins of extra utility here. The main obstacle to your initial experience was that we had failed to generate the release notes adjacent to the package you trialed!

dchadwic-xilinx: Thats likely true. If you are happy with the current order, then that is fine I think.

dchadwic-xilinx (approved):

doc/ReleaseNotes
abower-amd: Thanks Daniel!

PR #50, opened 2024-10-15
ON-16111: use monitor thread in zfsink to allow testing higher rates

rdrinkwa-xilinx:
Modified zfsink to use a monitoring thread to output traffic rate every second, as with efsink in Onload.

Tested with high packet rate (>5Mpps) and high throughput (>9Gbps) to confirm output is correctly formatted.
Timestamps work as before and are output every packet.
Monitoring thread can be disabled using '-q'.

### DIFF
--- a/src/tests/zf_apps/zfsink.c
+++ b/src/tests/zf_apps/zfsink.c
@@ -21,10 +21,23 @@
 #include <stdbool.h>
 #include <stdarg.h>
 #include <inttypes.h>
+#include <pthread.h>
+#include <sys/time.h>
+
+
+struct resources {
+  /* statistics */
+  volatile uint64_t n_rx_pkts;
+  volatile uint64_t n_rx_bytes;
+};
 
 
 static bool cfg_quiet = false;
 static bool cfg_rx_timestamping = false;
+static struct resources res;
+
+/* Mutex to protect printing from different threads */
+static pthread_mutex_t printf_mutex;
 
 
 static void usage_msg(FILE* f)
@@ -51,12 +64,12 @@ static void usage_err(void)
 
 static void vlog(const char* fmt, ...)
 {
-  if( ! cfg_quiet ) {
-    va_list args;
-    va_start(args, fmt);
-    vprintf(fmt, args);
-    va_end(args);
-  }
+  va_list args;
+  va_start(args, fmt);
+  pthread_mutex_lock(&printf_mutex);
+  vprintf(fmt, args);
+  pthread_mutex_unlock(&printf_mutex);
+  va_end(args);
 }
 
 
@@ -77,9 +90,6 @@ static void try_recv(struct zfur* ur)
 
     /* Do something useful with the datagram here! */
 
-
-    vlog("Received datagram of length %zu\n", rd.iov[0].iov_len);
-
     /* In the case rx timestamping capabilities are enabled, we can retrieve
      * the time at which the packet was received.
      * */
@@ -89,12 +99,15 @@ static void try_recv(struct zfur* ur)
       int rc = zfur_pkt_get_timestamp(ur, &rd.msg, &ts, 0, &flags);
 
       if( rc == 0 )
-        vlog("At time: %lld.%.9ld\n", ts.tv_sec, ts.tv_nsec);
+        vlog("Hardware timestamp: %lld.%.9ld\n", ts.tv_sec, ts.tv_nsec);
       else
         vlog("Error retrieving timestamp! Return code: %d\n", rc);
     }
 
     zfur_zc_recv_done(ur, &rd.msg);
+
+    res.n_rx_pkts += 1;
+    res.n_rx_bytes += rd.iov[0].iov_len;
   } while( rd.msg.dgrams_left );
 }
 
@@ -104,7 +117,6 @@ static void poll_muxer(struct zf_muxer_set* muxer, int timeout)
   struct epoll_event evs[8];
   const int max_evs = sizeof(evs) / sizeof(evs[0]);
 
-  vlog("Polling muxer\n");
   int n_ev = zf_muxer_wait(muxer, evs, max_evs, timeout);
 
   for( int i = 0; i < n_ev; ++i )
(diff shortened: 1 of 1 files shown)


pemberso-xilinx (commented):
The only data used by the monitor thread is n_rx_pkts and n_rx_bytes but you've created a data structure with a lot more in it.

I think this update would be possible with more minimal changes to the app if just n_rx_pkts and n_rx_bytes were shared.

Could you also add some examples of the output from testing?

rdrinkwa-xilinx:
I refactored the code to just pass the resource structure between functions because this felt cleaner to me and was also the approach used for `efsink`.

Result from testing different options:
```
[server1:static]$ ./zfsink 224.1.2.3:12345 # two bursts of 100 packets
# pkt-rate  bandwidth(Mbps)       total-pkts
         0                0                 0
         0                0                 0
       100                0               100
         0                0               100
       100                0               200
         0                0               200
^C
[server1:static]$ ./zfsink -m 224.1.2.3:12345 # two bursts of 100 packets
# pkt-rate  bandwidth(Mbps)       total-pkts
         0                0                 0
         0                0                 0
       100                0               100
       100                0               200
         0                0               200
^C
[server1:static]$ ./zfsink -w 224.1.2.3:12345 # two bursts of 100 packets
# pkt-rate  bandwidth(Mbps)       total-pkts
         0                0                 0
         0                0                 0
       100                0               100
       100                0               200
         0                0               200
         0                0               200
^C
[server1:static]$ ./zfsink -r 224.1.2.3:12345 # one burst of 5 packets
# pkt-rate  bandwidth(Mbps)       total-pkts
         0                0                 0
         0                0                 0
Hardware timestamp: 1729093248.149723774
Hardware timestamp: 1729093248.149724426
Hardware timestamp: 1729093248.149724499
Hardware timestamp: 1729093248.149724575
Hardware timestamp: 1729093248.149724651
         5                0                 5
         0                0                 5
         0                0                 5
^C
[server1:static]$ ./zfsink -qr 224.1.2.3:12345 # one burst of 5 packets
Hardware timestamp: 1729093258.533324537
Hardware timestamp: 1729093258.533325183
Hardware timestamp: 1729093258.533325258
Hardware timestamp: 1729093258.533325333
Hardware timestamp: 1729093258.533325410
^C
[server1:static]$ ./zfsink 224.1.2.3:12345 # efsend using 1 byte payload no throttling
# pkt-rate  bandwidth(Mbps)      total-pkts
   3534625               28          3534625
   5248144               41          8782769
   5247216               41         14029985
   5248816               41         19278801
   5248128               41         24526929
^C
[server1:static]$
```

pemberso-xilinx:
Thanks for the output, looks fine.

> I refactored the code to just pass the resource structure between functions because this felt cleaner to me and was also the approach used for efsink.

efsink needs to manage packet buffers so there is more need to share state.  I'd rather have a minimal code change here - e.g. suggest just adding a couple of global stats counters and then most of the changes to existing code won't be needed.  Code can always be refactored later if more data needs to be shared but doesn't seem necessary here.

src/tests/zf_apps/zfsink.c
```diff
@@ -233,8 +247,51 @@ void print_attrs(struct zf_attr* attr)
...
+  while( 1 ) {
+    sleep(1);
+    now_pkts = res->n_rx_pkts;
+    now_bytes = res->n_rx_bytes;
+    gettimeofday(&end, NULL);
+    ms = (end.tv_sec - start.tv_sec) * 1000;
+    ms += (end.tv_usec - start.tv_usec) / 1000;
+    pkt_rate = (int) ((now_pkts - prev_pkts) * 1000 / ms);
+    mbps = (int) ((now_bytes - prev_bytes) * 8 / 1000 / ms);
+    pthread_mutex_lock(&printf_mutex);
```
ligallag-amd, line 272: I don't think a mutex is needed for printf. I believe it synchronises itself at least in my experiments.

src/tests/zf_apps/zfsink.c
```diff
@@ -233,8 +247,51 @@ void print_attrs(struct zf_attr* attr)
...
+  printf("#%9s %16s %16s", "pkt-rate", "bandwidth(Mbps)", "total-pkts\n");
+
+  prev_pkts = res->n_rx_pkts;
+  prev_bytes = res->n_rx_bytes;
+  gettimeofday(&start, NULL);
+
+  while( 1 ) {
+    sleep(1);
+    now_pkts = res->n_rx_pkts;
+    now_bytes = res->n_rx_bytes;
```
ligallag-amd, line 266: Also, because it's shared data, res should at least be volatile. If you want to be completely correct, you should use [atomic builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) with relaxed ordering.

src/tests/zf_apps/zfsink.c
```diff
@@ -290,6 +347,8 @@ int main(int argc, char* argv[])
   if( cfg_rx_timestamping )
     ZF_TRY(zf_attr_set_int(attr, "rx_timestamping", 1));
 
+  ZF_TRY((res = calloc(1, sizeof(*res))) != NULL);
```
pemberso-xilinx, line 350: Given this is just a small test app, I think we could just have a single global instance of struct resources rather than dynamically allocating and passing around the functions.

pemberso-xilinx (changes requested):

ligallag-amd (approved):
Looks good to me

src/tests/zf_apps/zfsink.c
```diff
@@ -21,10 +21,23 @@
...
+struct resources {
+  /* statistics */
+  volatile uint64_t n_rx_pkts;
+  volatile uint64_t n_rx_bytes;
+};
 
 
 static bool cfg_quiet = false;
 static bool cfg_rx_timestamping = false;
+static struct resources res;
```
pemberso-xilinx, line 37: minor comment - I would use

`static struct resources res = {0};`

here and save the lines initialising later in `main()`

tcrawley-xilinx: Objects with `static` storage duration will [implicitly](https://en.cppreference.com/w/c/language/initialization#:~:text=objects%20with%20static%20and%20thread%2Dlocal%20storage%20duration%20are%20empty%2Dinitialized) be [empty-initialised](https://en.cppreference.com/w/c/language/initialization#:~:text=pointers%20are%20initialized,initialized%20to%20zero), so the current declaration is sufficient to set the values to 0.

PR #51, opened 2024-10-25
ON-16143: remove -p option from zfsink test application. zf_stackdump attr can be used as alternative. 

KalepuAMD:
ON-15038: View zf attibutes from zf_stackdump. The Pull request for this enhancement has been merged. 
As zfsink has a function to printout the zf attributes; removing it from application to reduce the volume of the code. 
Workaround for this is to use zf_stackdump attr for getting the attributes

Test the application and all working fine.

### DIFF
--- a/src/tests/zf_apps/zfsink.c
+++ b/src/tests/zf_apps/zfsink.c
@@ -38,7 +38,6 @@ static void usage_msg(FILE* f)
   fprintf(f, "  -w       Use the zf waitable fd\n");
   fprintf(f, "  -r       Enable rx timestamping\n");
   fprintf(f, "  -q       Quiet -- do not emit progress messages\n");
-  fprintf(f, "  -p       Print zf attributes after stack startup\n");
 }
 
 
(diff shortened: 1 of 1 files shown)


ligallag-amd (approved):

ligallag-amd:
> I think I would probably want this command line option to stick around even though it has been superceded by other functionality. 

I don't think anything uses this option. I had a look in smartnic-runbench and the single test for zfsink doesn't use it. What else might? I don't see a major reason to get rid of this code though, this is just a test-app after all.

jfeather-amd:
> > I think I would probably want this command line option to stick around even though it has been superceded by other functionality.
> 
> I don't think anything uses this option. I had a look in smartnic-runbench and the single test for zfsink doesn't use it. What else might? I don't see a major reason to get rid of this code though, this is just a test-app after all.

It's reassuring to know that no tests seem to use this option (as even my suggestion would probably need rewriting of any output parsing), but I don't think there's any reason to justify removing it entirely. I would understand wanting to deduplicate the functionality by changing this code to use the `zf_stackdump` functions, but entirely removing this seems a bit unnecessary to me.

Also worth noting I left my previous comment as not a request for changes, so would be happy with this being merged if anyone disagrees. Perhaps worth also including @matthewr-xilinx who (as far as I can tell) originally suggested it might be worth removing this option after ON-15038 added `zf_stackdump attr`.

src/tests/zf_apps/zfsink.c
```diff
@@ -160,84 +159,10 @@ static void ev_loop_waitable_fd(struct zf_stack* stack,
...
-
-
 int main(int argc, char* argv[])
 {
   int cfg_muxer = 0;
   int cfg_waitable_fd = 0;
-  bool cfg_print_attrs = false;
   
   int c;
   while( (c = getopt(argc, argv, "hmrwqp")) != -1 )
```
matthewr-xilinx, line 168: Ought to remove `p` from here too?

matthewr-xilinx (approved):
On balance, I think I'm happy to reduce the length/complexity of zfsink by removing this option. It was mostly added because at that time we didn't have the `zf_stackdump attr` option.

In theory we could keep this option as a demonstration of how to use `zf_attr_get_int()`/`zf_attr_doc()` etc. However I don't think many people have actually needed this.

Discussed with @KalepuAMD and proposed updating the commit message (to explicitly say that this commit removes the `-p` option, and that `zf_stackdump attr` can be used instead).

src/tests/zf_apps/zfsink.c
KalepuAMD: Yes, it is only used for print attributes. as we removed the print attributes function; i removed it

PR #52, opened 2024-10-30
ON-16037: added functions to read stats for interfaces

rdrinkwa-xilinx:
Added two functions to read the drop stats from the physical interfaces used by TCPDirect. Tested as working with physical interfaces themselves or a bond. The zfsink app was updated to make use of this additional functionality for testing and demo purposes.

The design mimics that of ef_vi itself with the two calls to get the layout and then the actual stats themselves. It could be changed to hold a copy of the layout internally within the stack so passing it back in wouldn't be required but it would add complexity and necessitate code changes in other files beyond the minimal updates to existing files here.

Testing with 'efsend' sending 1000 1-byte packets as a burst that is known to exceed zfsink's ability to avoid drops.
```
# testing physical interface with regular drops
[server1:static]$ ZF_ATTR="interface=enp1s0f0" ./zfsink -s 224.1.2.3:12345
# pkt-rate  bandwidth(Mbps)       total-pkts  RX CRC errors  RX trunk errors  RX no descriptor errors  RX abort errors
         0                0                0              0                0                        0                0
         0                0                0              0                0                        0                0
       710                0              710              0                0                      290                0
         0                0              710              0                0                        0                0
         0                0              710              0                0                        0                0
^C
# physical interface and backgrounded to cause more drops
[server1:static]$ ZF_ATTR="interface=enp1s0f0" ./zfsink -s 224.1.2.3:12345
# pkt-rate  bandwidth(Mbps)       total-pkts  RX CRC errors  RX trunk errors  RX no descriptor errors  RX abort errors
         0                0                0              0                0                        0                0
^Z
[1]+  Stopped                 ./zfsink -s 224.1.2.3:12345
[server1:static]$ fg
ZF_ATTR="interface=enp1s0f0" ./zfsink -s 224.1.2.3:12345
         0                0                0              0                0                      496                0
       504                0              504              0                0                        0                0
         0                0              504              0                0                        0                0
^C
 # running on a bond with burst on first then second intf. Output changed to show one line per intf.
[server1:static]$ ZF_ATTR="interface=bond0" ./zfsink -s 224.1.2.3:12345
# pkt-rate  bandwidth(Mbps)       total-pkts
#  RX CRC errors RX trunk errors RX no descriptor errors RX abort errors
              0                0                        0                0
              0                0                        0                0
              0                0                        0                0
              0                0                        0                0
       744                0              744
              0                0                      256                0
              0                0                        0                0
         0                0              744
              0                0                        0                0
              0                0                        0                0
       760                0             1504
              0                0                        0                0
              0                0                      240                0
         0                0             1504
              0                0                        0                0
              0                0                        0                0
         0                0             1504
              0                0                        0                0
              0                0                        0                0
^C
[server1:static]$

```


src/include/zf/zf_stack.h
```diff
@@ -167,5 +167,71 @@ enum zf_stack_feature {
...
+
+/*! \brief Layout for statistics */
+typedef struct {
+  /** Size of data for statistics */
+  int                      zfsl_data_size;
+  /** Number of fields of statistics */
+  int                      zfsl_fields_num;
+  /** Array of fields of statistics */
+  zf_stats_field_layout    zfsl_fields[];
+} zf_stats_layout;
```
ligallag-amd, line 188: I know this will be quite laborious but these structs absolutely need to have the exact same layout as their ef_vi counterparts. Can you add some build asserts to ensure the offsets and sizes of each field are the same?

src/tests/zf_apps/zfsink.c
```diff
@@ -243,15 +244,81 @@ void print_attrs(struct zf_attr* attr)
...
+
+
+static void monitor(struct zf_stack* stack)
 {
   uint64_t now_bytes, prev_bytes;
   struct timeval start, end;
   uint64_t prev_pkts, now_pkts;
   int ms, pkt_rate, mbps;
 
+  const zf_stats_layout* zf_stats_layout[4];
```
ligallag-amd, line 305: Is there any way we can get rid of hard coding 4? is there a way to deduce this length?

ligallag-amd (changes requested):
I think you need to make static method which does the following.
``` c
pthread_mutex_lock(&printf_mutex);
printf(...)
pthread_mutex_unlock(&printf_mutex);
```
There is too much duplicated code. Please update https://github.com/Xilinx-CNS/tcpdirect/pull/50 to do this also.

src/tests/zf_apps/zfsink.c
rdrinkwa-xilinx: I've added a function that will report how many interfaces there are for a given TCPDirect stack. This will return 1 unless it's a bond when it will return the number of underlying interfaces.

src/lib/zf/zf_stats.c
```diff
@@ -0,0 +1,69 @@
...
+               "zf_stats_field_layout member evsfl_name does not match ef_vi_stats_field_layout");
+static_assert( offsetof(zf_stats_layout, evsl_fields_num) == offsetof(ef_vi_stats_layout, evsl_fields_num),
+               "zf_stats_field_layout member evsfl_name does not match ef_vi_stats_field_layout");
+static_assert( offsetof(zf_stats_layout, evsl_fields) == offsetof(ef_vi_stats_layout, evsl_fields),
+               "zf_stats_field_layout member evsfl_name does not match ef_vi_stats_field_layout");
+
+
+int
+zf_stats_query_layout(struct zf_stack* stack,
+                      const zf_stats_layout** const layout_out, int layout_sz)
```
pemberso-xilinx, line 28: Rather than have the user get interface count and malloc, I would consider making this zf_alloc_stats_query_layout() and return a new data type with layout(s) and size.  Would need a corresponding zf_free_stats_query_layout()

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,80 @@ void print_attrs(struct zf_attr* attr)
...
+  zf_stats_layout = malloc(nic_cnt * sizeof(struct zf_stats_layout *));
+  ZF_TEST(zf_stats_layout != NULL);
+
+  pthread_mutex_lock(&printf_mutex);
+  printf("#%9s %16s %16s", "pkt-rate", "bandwidth(Mbps)", "total-pkts");
+
+  if( cfg_intf_stats ) {
+    nic_cnt = zf_stats_header_print(stack, nic_cnt, zf_stats_layout);
+    for( i = 0; i < nic_cnt; i++ )
+      malloc_sz += zf_stats_layout[i]->evsl_data_size;
```
pemberso-xilinx, line 316: rather than making the user do this, I'd have a utility to allocate data sufficient for new data structure containing layouts mentioned above

src/tests/zf_apps/zfsink.c
```diff
@@ -265,8 +333,17 @@ static void monitor()
     ms += (end.tv_usec - start.tv_usec) / 1000;
     pkt_rate = (int) ((now_pkts - prev_pkts) * 1000 / ms);
     mbps = (int) ((now_bytes - prev_bytes) * 8 / 1000 / ms);
-    vlog("%10d %16d %16"PRIu64"\n", pkt_rate, mbps, now_pkts);
+    
+    pthread_mutex_lock(&printf_mutex);
+    printf("%10d %16d %16"PRIu64, pkt_rate, mbps, now_pkts);
+
+    if( cfg_intf_stats )
+      zf_stats_print(stack, 1, stats_data, nic_cnt, zf_stats_layout);
```
pemberso-xilinx, line 341: I was confused by passing empty stats data into a function name ..._print().  I would pull call to get the stats here before the print.

pemberso-xilinx (changes requested):

src/lib/zf/zf_stats.c
```diff
@@ -0,0 +1,69 @@
+/* SPDX-License-Identifier: MIT */
+/* SPDX-FileCopyrightText: (c) Advanced Micro Devices, Inc. */
+#include <zf_internal/zf_stack.h>
+
+/* ensure ef_vi/zf structures are the same size in case
```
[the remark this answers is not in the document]
rdrinkwa-xilinx: When it's used with ef_vi_transmit_alt_overhead that is because the user application will be making an ef_vi call. That's not the case here and I wanted to avoid having a user application pull in vi.h just to get two structure definitions.

src/lib/zf/zf_stats.c
pemberso-xilinx: My personal preference is to have a specific free call to make it symmetric. Makes code easier to follow (IMO) and in theory the data structure could contain pointers to other objects that needed to be freed first in future. (I'd also advise taking a look in the library for any precedents one way or the other).

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,75 @@ void print_attrs(struct zf_attr* attr)
...
+  int i, malloc_sz = 0;
+
+  pthread_mutex_lock(&printf_mutex);
+  printf("#%9s %16s %16s", "pkt-rate", "bandwidth(Mbps)", "total-pkts");
+
+  if( cfg_intf_stats ) {
+    ZF_TEST(zf_stats_alloc_query_layout(stack, &stats_collection) == 0);
+    zf_stats_header_print(stack, stats_collection);
+
+    for( i = 0; i < stats_collection->num_intfs; i++ )
```
pemberso-xilinx, line 310: could we include a field in the collection (e.g. total_data_size) to avoid the user needing this?

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,75 @@ void print_attrs(struct zf_attr* attr)
 }
 
 
-static void monitor()
+static void zf_stats_header_print(struct zf_stack* stack,
+                                  zf_stats_collection* stats_collection)
+{
+  int i;
+
+  /* only display names for the first NIC and assume others are the same */
```
pemberso-xilinx, line 253: this comment doesn't fit with the code immediately below it that is printing a new line for multiple NIC case

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,75 @@ void print_attrs(struct zf_attr* attr)
...
+  if( stats_collection->num_intfs > 1 ) {
```
pemberso-xilinx, line 254: coding style doesn't use braces for if condition containing single statement

src/include/zf/zf_stack.h
```diff
@@ -167,5 +167,81 @@ enum zf_stack_feature {
 */
 ZF_LIBENTRY int zf_stack_query_feature(struct zf_stack* stack, enum zf_stack_feature feature);
 
+/*! \brief Layout for a field of statistics */
```
pemberso-xilinx, line 170: not sure zf_stack.h is correct place for all this.  I would consider new zf_stats.h (and add to zf.h)

src/include/zf/zf_stack.h
```diff
@@ -167,5 +167,81 @@ enum zf_stack_feature {
...
+typedef struct {
+  /** Name of statistics field */
+  char* evsfl_name;
+  /** Offset of statistics file, in bytes */
```
pemberso-xilinx, line 174: file => field?

src/include/zf/zf_stack.h
```diff
@@ -167,5 +167,81 @@ enum zf_stack_feature {
...
+  zf_stats_field_layout    evsl_fields[];
+} zf_stats_layout;
+
+/*! \brief  */
+typedef struct {
+  /** Number of underlying interfaces present in stack */
+  int               num_intfs;
+  /** Array of layouts, one per interface */
+  zf_stats_layout** layout;
+} zf_stats_collection;
```
pemberso-xilinx, line 196: Think this should have layout in name as doesn't contains stats.
could perhaps just call this zf_stats_layout and name the data structures above zf_nic_stats_...

src/include/zf/zf_stack.h
```diff
@@ -167,5 +167,81 @@ enum zf_stack_feature {
...
+** \param collection Pointer to an zf_stats_collection, that is allocated and
+**                   updated on return with the layout for available
+**                   statistics. This must be released when no longer needed
+**                   using zf_stats_free_query_layout().
+**
+** \return Zero on success or negative error code.
+**
+** Retrieve layout for available statistics.
+*/
+ZF_LIBENTRY int zf_stats_alloc_query_layout(struct zf_stack* stack,
```
pemberso-xilinx, line 210: I would name this and _free functions in relation to data structure they allocate.  See also note above on data structure name.

src/tests/zf_apps/zfsink.c
```diff
@@ -265,18 +328,34 @@ static void monitor()
     ms += (end.tv_usec - start.tv_usec) / 1000;
     pkt_rate = (int) ((now_pkts - prev_pkts) * 1000 / ms);
     mbps = (int) ((now_bytes - prev_bytes) * 8 / 1000 / ms);
-    vlog("%10d %16d %16"PRIu64"\n", pkt_rate, mbps, now_pkts);
+    
+    pthread_mutex_lock(&printf_mutex);
+    printf("%10d %16d %16"PRIu64, pkt_rate, mbps, now_pkts);
+
+    if( cfg_intf_stats ) {
+      zf_stats_query(stack, stats_data, stats_collection, 1);
```
pemberso-xilinx, line 336: It would be cleaner to move the zf_stats_query outside of the printf lock as it doesn't need it

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,75 @@ void print_attrs(struct zf_attr* attr)
...
+  zf_stats_collection* stats_collection;
+  uint8_t* stats_data = NULL;
+  int i, malloc_sz = 0;
+
+  pthread_mutex_lock(&printf_mutex);
+  printf("#%9s %16s %16s", "pkt-rate", "bandwidth(Mbps)", "total-pkts");
+
+  if( cfg_intf_stats ) {
+    ZF_TEST(zf_stats_alloc_query_layout(stack, &stats_collection) == 0);
+    zf_stats_header_print(stack, stats_collection);
```
pemberso-xilinx, line 308: only this zf_stats_head_print is using the printf lock so would move other code before it

pemberso-xilinx (changes requested):

src/tests/zf_apps/zfsink.c
rdrinkwa-xilinx: There is a new line printed but the names output are only for the first NIC, hence `layout[0]` being used.

src/include/zf/zf_stack.h
rdrinkwa-xilinx: The `zf_stats_field_layout` and `zf_stats_layout` relate directly to the `ef_vi_...` ones of the same name. I've updated this to `zf_layout_collection`, since that is what it's mostly used for.

src/tests/zf_apps/zfsink.c
pemberso-xilinx: Moving comment down to that code would be clearer.

src/tests/zf_apps/zfsink.c
```diff
@@ -243,14 +245,75 @@ void print_attrs(struct zf_attr* attr)
...
+static void zf_stats_print(struct zf_stack* stack, uint8_t* stats_data,
+                           zf_stats_collection* stats_collection)
+{
+  int i, n, n_pad;
+  uint8_t* cur_data = stats_data;
+  const zf_stats_layout * layout;
+
+  for( n = 0; n < stats_collection->num_intfs; n++ ) {
+    if( stats_collection->num_intfs > 1 )
+      printf("\n");
```
pemberso-xilinx, line 271: I think this needs to be "\n " (added space) to account for "#" in headers.  Looking at example output, values weren't aligned for multiple NIC case.

src/lib/zf/zf_stats.c
```diff
@@ -0,0 +1,91 @@
...
+
+int
+zf_stats_query(struct zf_stack* stack, void* data,
+               zf_layout_collection* collection, int do_reset)
+{
+  int i, rc = 0;
+  int num_intfs = stack->nics_n;
+  char* pData = (char*)data;
+
+  assert(num_intfs == collection->num_intfs);
```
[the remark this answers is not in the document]
rdrinkwa-xilinx: This is needed for users to know how many sets of stats there are present. The 'efsink' example demonstrates how this can be used.

ligallag-amd (approved):

src/include/zf/zf_stats.h
```diff
@@ -0,0 +1,77 @@
...
+
+/*! \brief  */
+typedef struct {
+  /** Number of underlying interfaces present in stack */
+  int               num_intfs;
+  /** Size of memory needed to query the stats in bytes */
+  int               total_data_size;
+  /** Array of layouts, one per interface */
+  zf_stats_layout** layout;
+} zf_layout_collection;
```
pemberso-xilinx, line 29: The fact this doesn't include stats in the name bothers me slightly. Wondering if you could name it `zf_vi_stats` (for consistency with onload) or possibly `zf_nic_stats` or `zf_interface_stats`  with appropriate renaming elsewhere (e.g. `zf_alloc_vi_stats`)?

PR #58, opened 2024-12-17
ON-16235: refactor GHA to be reusable and add unit test options

jfeather-amd:
The motivation behind this patch series is to allow the TCPDirect GitHub Actions to be run from other repositories, specifically onload. This is useful to make sure onload changes don't break the build/tests of TCPDirect.

The first commit is the main work here, which extracts actions (install dependencies, build, and test) from the build workflow and places them into their own individual action files which can be called independently. The remaning commits work around issues caused by private repos having less powerful runners. The second commit, to add a timeout multiplier, is actually more for myself as my dev machine is too slow to pass some of the superbuf tests - filtering out slow tests is sufficient to pass GHA on slower runners.

### Testing Done
- [TCPDirect GHA is effectively a no-op](https://github.com/jfeather-amd/tcpdirect/actions/runs/12370377132/job/34524216075)
- My Jenkins personal pipeline runs all tests now (rather than just the unstable ones)
- `make test` with a variety of `ZF_RUN_SLOW_TESTS` and `ZF_RUN_UNSTABLE_TESTS`

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) 2024 Advanced Micro Devices, Inc.
+
+name: Build
+description: Build TCPDirect
+inputs:
+  tcpdirect_tree:
+    required: true
+    type: string
+    description: Path to the TCPDirect source tree to build
+  onload_tree:
+    required: true
+    type: string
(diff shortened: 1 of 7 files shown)


ligallag-amd (approved):
LGTM after a quick scan through

jfeather-amd:
Latest push incorporates review comments and fixes an issue in input validation for the build action. Planning on merging this when the actions on this branch finish running (and pass)

PR #62, opened 2025-01-16
ON-16265: Set no-poison vi flag when ctpio_mode=sf-np

dchadwic-xilinx:
Since `ctpio_mode` can be <0, using a signed int is more appropriate. (https://github.com/Xilinx-CNS/tcpdirect/blob/tcpdirect-9.0.0/src/lib/zf/private/stack_alloc.c#L52-L54)

We also decided that there is no obvious reason for the no-poison flag to be set in the event that `ctpio_mode < 0`. My assumption is that there was an earlier decision that any future 'no-poisoning' CTPIO modes would also have negative values like `CTPIO_MODE_SF_NP=-1`.

Another solution would be to use an enum but I do not see examples of this being used elsewhere in this codebase. This solution is the least disruptive imo.

As a test, I created a `sf-np` VI and then a `ct` VI (on same interface) and strated tx'mitting with the `ct` VI.
- Base v8_1: saw `rx_bad` accumulate on remote end.
- v8_1 + this patch: no poisoned frames hit the wire.

### DIFF
--- a/src/lib/zf/private/stack_alloc.c
+++ b/src/lib/zf/private/stack_alloc.c
@@ -290,7 +290,7 @@ static int zf_stack_init_pio(struct zf_stack_impl* sti, struct zf_attr* attr,
 
 static int zf_stack_init_ctpio_stack_config(struct zf_stack_impl* sti,
                                             struct zf_attr* attr,
-                                            unsigned* ctpio_mode)
+                                            int* ctpio_mode)
 {
   zf_stack* st = &sti->st;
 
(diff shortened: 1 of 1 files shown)


jfeather-amd (commented):
Looks reasonable, have you done any testing? It would be nice to see if you could force CTPIO poisoning with `CTPIO_MODE_SF` then show that no poisoning happens with `CTPIO_MODE_SF_NP` although I don't know how easy/plausible that would be to do.

dchadwic-xilinx:
Testing was done here: proving that the vi_flags are now showing correctly 3000000 rather than 1000000. I'll have a go at forcing poisoning.

jfeather-amd:
> I had a spare moment so proved the fix does correct the behaviour

Nice, thanks for doing this!

PR #68, opened 2025-02-18
ON-14783: Use additional NIC flags for X3522

ibeecraft-amd:
Running with Warnings enabled on X3 you get output warning that the NIC is not running low-latency firmware. 

This commit simply suppresses the warning on X3 NICs.

Without commit on X3522:
```
ZF_ATTR="log_level=0x2;interface=ens1f0np0" ./zfsink 224.1.1.1:9001
  1000 | Interface ens1f0np0 is not in low latency mode.
  1000 | Low latency mode is recommended for best latency with TCPDirect.
  1000 | PIO not supported by efct interface but pio=3. Not attempting to allocate PIO buffer.
Polling reactor
```

With commit on X3522:
```
ZF_ATTR="log_level=0x2;interface=ens1f0np0" ./zfsink 224.1.1.1:9001
  1000 | PIO not supported by efct interface but pio=3. Not attempting to allocate PIO buffer.
Polling reactor
```
With Commit on X2522 running full-feature firmware:
```
ZF_ATTR="log_level=0x2;interface=enp131s0f0" ./zfsink 224.1.1.1:9001
  1000 | Interface enp131s0f0 is not in low latency mode.
  1000 | Low latency mode is recommended for best latency with TCPDirect.
Polling reactor
```

With commit on X2522 running ultra-low-latency firmware:
```
ZF_ATTR="log_level=0x2;interface=enp131s0f0" ./zfsink 224.1.1.1:^C01
sudo ip l set enp129s0f0 up
ZF_ATTR="log_level=0x2;interface=enp129s0f0" ./zfsink 224.1.1.1:9001
Polling reactor
```

### DIFF
--- a/src/lib/zf/private/stack_alloc.c
+++ b/src/lib/zf/private/stack_alloc.c
@@ -12,6 +12,7 @@
 \**************************************************************************/
 
 
+#include <cerrno>
 #include <zf/zf.h>
 #include <zf_internal/private/zf_hal.h>
 #include <zf_internal/zf_pool_res.h>
@@ -242,19 +243,25 @@ static int zf_stack_init_pio(struct zf_stack_impl* sti, struct zf_attr* attr,
   struct zf_stack_nic* st_nic = &st->nic[nicno];
   struct zf_stack_res_nic* sti_nic = &sti->nic[nicno];
   ef_vi* vi = zf_stack_nic_tx_vi(st_nic);
-
-  if( st_nic->vi.nic_type.arch == EF_VI_ARCH_EFCT &&
-      attr->pio >= PIO_MUST_USE ) {
-    zf_log_stack_warn(st,
-                      "PIO not supported by efct interface but pio=%d. "
-                      "Not attempting to allocate PIO buffer.\n",
-                      attr->pio);
-    st_nic->pio.busy = 3;
-    return 0;
+  unsigned long pio_available;
+  /* Check whether the NIC supports PIO mode. */
+  int rc = ef_pd_capabilities_get(sti_nic->dh, &sti_nic->pd, sti_nic->dh,
+                                  EF_VI_CAP_PIO, &pio_available);
+  if ( rc != 0 ) {
+    pio_available = 0;
+    zf_log_stack_warn(st, "Failed to query PIO capability (rc = %d)\n", rc);
+    if( attr->pio >= PIO_MUST_USE ) {
+      zf_log_stack_warn(st,
+                        "PIO not supported by %s but pio=%d. "
+                        "Not attempting to allocate PIO buffer.\n",
+                        sti_nic->if_name, attr->pio);
+      st_nic->pio.busy = 3;
+      return 0;
+    }
   }
 
   /* Try to allocate. If alloc fails and we are in-strict mode (2) then fail */
-  int rc = ef_pio_alloc(&sti_nic->pio, sti_nic->dh, &sti_nic->pd,
+  rc = ef_pio_alloc(&sti_nic->pio, sti_nic->dh, &sti_nic->pd,
                         -1, sti_nic->dh);
   if( rc < 0 && attr->pio >= PIO_MUST_USE ) {
     zf_log_stack_err(st, "Failed to allocate PIO buffer (rc = %d)\n", rc);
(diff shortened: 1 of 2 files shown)


src/lib/zf/private/stack_alloc.c
```diff
@@ -654,6 +646,15 @@ int zf_stack_init_nic_resources(struct zf_stack_impl* sti,
     ef_vi_receive_set_buffer_len(&st_nic->vi, PKT_BUF_SIZE_USABLE);
   }
 
+  if( (!(sti->nic[nicno].flags & ZF_RES_NIC_FLAG_RX_LL) ||
+      !(sti->nic[nicno].flags & ZF_RES_NIC_FLAG_TX_LL)) &&
+      !(st_nic->vi.nic_type.arch == EF_VI_ARCH_EFCT) ) {
```
sianj-xilinx, line 651: Don't test on arch, use the capabilities API. If your question can't be answered with an existing capability we should add a new one.

jfeather-amd, line 651: It would be nice if we had a way to communicate that the capability check actually did something meaningful instead of just [directly returning `nic->rx_variant`](https://github.com/Xilinx-CNS/onload/blob/bf581258c7d73704a090ea43cf1f5d2a8d2d2868/src/driver/linux_char/efch_capabilities.c#L293-L295) and ideally plumb that information through from [where we get this data](https://github.com/Xilinx-CNS/tcpdirect/blob/14d93228f700fad32ea7fa93f88d79d59d7748ad/src/lib/zf/private/stack_alloc.c#L502). Not sure how feasible this is, but at the very least, could this instead check for `EF_VI_ARCH_EF10`? It seems that's the only architecture [that sets `rx_variant`](https://github.com/Xilinx-CNS/onload/blob/bf581258c7d73704a090ea43cf1f5d2a8d2d2868/src/lib/efhw/ef10/ef10.c#L510) to a value that makes sense to check here (this also seems to be true for the head of onload-8.1 as far as I can see).

sianj-xilinx: I agree that a reasonable approach would be to make the FW variant capability be more nuanced in its reporting, so that it only reported success on NICs that support multiple variants. That should be pretty simple. It's technically a change in behaviour, but I think it's justified.

ibeecraft-amd: I did consider this but the X3 does not report useful information from the GET CAPABILITIES MCDI command for referencing the firmware variant. It is possible to just write the value into the ef_vi API instead. If that is preferable? 

I will close this PR and raise one onto Onload for working around this issue.

ibeecraft-amd: Is there a reason why we can't test on the arch. I only ask because TCPDirect is littered with similar test, including in this same function.

sianj-xilinx: It shouldn't be. We should remove them, but in the meantime we shouldn't add more. It's a problem because it makes code harder to update when we implement new things. The upper layers of onload and tcpdirect should be hardware agnostic.

sianj-xilinx: The capabilities API is not a direct mapping of the MCDI command - AF_XDP interfaces don't support MCDI at all, and various things we report don't come from the GET_CAPS command, so lack of support over MCDI is not an issue. We just need all arches to understand what they support, and report it appropriately.

src/lib/zf/private/stack_alloc.c
```diff
@@ -243,12 +243,12 @@ static int zf_stack_init_pio(struct zf_stack_impl* sti, struct zf_attr* attr,
   struct zf_stack_res_nic* sti_nic = &sti->nic[nicno];
   ef_vi* vi = zf_stack_nic_tx_vi(st_nic);
 
-  if( st_nic->vi.nic_type.arch == EF_VI_ARCH_EFCT &&
+  if( sti_nic->flags & ZF_RES_NIC_FLAG_CTPIO_ONLY &&
```
sianj-xilinx, line 246: This should instead be based on the whether the NIC has the capability `EF_VI_CAP_PIO`.

src/lib/zf/private/stack_alloc.c
```diff
@@ -492,15 +492,24 @@ static int zf_stack_init_nic_capabilities(struct zf_stack* st, int nicno)
...
+  if( rc == -EOPNOTSUPP ) {
+    rc = ef_pd_capabilities_get(dh, pd, dh, EF_VI_CAP_RX_SHARED, &variant2);
+    if( rc != 0 ) {
+      zf_log_stack_err(st, "Unknown RX mode for interface %s (rc %d)\n",
+                           sti->nic[nicno].if_name, rc);
+      return -EOPNOTSUPP;
+    }
+    sti->nic[nicno].flags |= ZF_RES_NIC_FLAG_SHARED_RXQ;
+  }
+  else if ( rc != 0 ) {
```
sianj-xilinx, line 512: I don't think it's reasonable to equate "shared RXQ" (ie X3) with not supporting report of FW variant. I think we can just downgrade the logging in the EOPNOTSUPP case to debug only.

src/lib/zf/private/stack_alloc.c
```diff
@@ -515,7 +524,16 @@ static int zf_stack_init_nic_capabilities(struct zf_stack* st, int nicno)
...
+  if( rc == -EOPNOTSUPP ) {
+    rc = ef_pd_capabilities_get(dh, pd, dh, EF_VI_CAP_CTPIO_ONLY, &variant2);
+    if( rc != 0 ) {
+      zf_log_stack_err(st, "Unknown TX mode for interface %s (rc %d)\n",
+                           sti->nic[nicno].if_name, rc);
+      return -EOPNOTSUPP;
+    }
+    sti->nic[nicno].flags |= ZF_RES_NIC_FLAG_CTPIO_ONLY;
+  }
+  else if( rc != 0 ) {
```
sianj-xilinx, line 536: The same here - we shouldn't conflate reporting of FW variant with other capabilities. These should be checked independently.

src/lib/zf/private/stack_alloc.c
```diff
@@ -636,9 +655,9 @@ int zf_stack_init_nic_resources(struct zf_stack_impl* sti,
   );
 
   if ( attr->rx_ring_max != 0 ) {
-    /* For EFCT, we store the timestamp in a fake prefix when copying from
-     * the shared rx buffer into our own packet buffer. */
-    if( st_nic->vi.nic_type.arch == EF_VI_ARCH_EFCT )
+    /* For shared rxqs, we store the timestamp in a fake prefix when copying
+     * from the shared rx buffer into our own packet buffer. */
+    if( sti_nic->flags & ZF_RES_NIC_FLAG_SHARED_RXQ )
       st_nic->rx_prefix_len = ES_DZ_RX_PREFIX_SIZE;
```
sianj-xilinx, line 661: This should be based on `EF_VI_CAP_RX_REF`.

src/lib/zf/private/stack_alloc.c
ibeecraft-amd: This check is to allow TCPDirect to function if PIO is requested but X3 NIC is present. It is a historic thing we have allowed but happy to remove this and error out instead but this would be a change to behaviour.

src/lib/zf/private/stack_alloc.c
ibeecraft-amd: Happy to make the change but this does not equate the two., it is an additional check in the error path.  I will re-jig the code to make it flow clearer.

It was a change to support X3 as it will fail the FW variant. Previous this has caused TCPDirect to exit so checking if it is a low latency VI through this RX shared capability check was to allow it to continue.

src/lib/zf/private/stack_alloc.c
ibeecraft-amd: This was based against Onload-9 branch which does not have this capability I can update it and rebase to master.

src/lib/zf/private/stack_alloc.c
sianj-xilinx: The documentation refers to warning if PIO not supported - I think the original check on EFCT arch was bad and this should always have been a check on the PIO capability.

src/lib/zf/private/stack_alloc.c
sianj-xilinx: This sets the SHARED_RXQ flag in exactly the case where reporting FW variant returns EOPNOTSUPP, which looks like equating the two to me. It's definitely not the right thing to do.

ibeecraft-amd: Right sorry I misunderstood at first. I see. Yes I will change this but I think I will leave this PR until some changes are made to the EF_VI capabilities API then rebase to master.

src/lib/zf/private/stack_alloc.c
```diff
@@ -242,19 +243,25 @@ static int zf_stack_init_pio(struct zf_stack_impl* sti, struct zf_attr* attr,
...
-                      attr->pio);
-    st_nic->pio.busy = 3;
-    return 0;
+  unsigned long pio_available;
+  /* Check whether the NIC supports PIO mode. */
+  int rc = ef_pd_capabilities_get(sti_nic->dh, &sti_nic->pd, sti_nic->dh,
+                                  EF_VI_CAP_PIO, &pio_available);
+  if ( rc != 0 ) {
+    pio_available = 0;
+    zf_log_stack_warn(st, "Failed to query PIO capability (rc = %d)\n", rc);
```
ligallag-amd, line 252: I'm confused in this case. You've set pio_available to 0, but we still try `ef_pio_alloc` if `!(attr->pio >= PIO_MUST_USE)`. I think we may as well exit here if `ef_pio_alloc` will always fail for x3 - which it should right? That said, I see that you're trying to preserve previous behaviour so maybe you can keep it as is

ibeecraft-amd: Yes sorry, I pushed it but hadn't finished working on this. Either way I see two ways of proceeding with this:

1. Remove the setting of pio_available as it is not used
2. Change the following code to use this new capability check (hesitant to do so as although should provide the same end result uses different function calls)

src/lib/zf/private/stack_alloc.c
ibeecraft-amd: If the following commit is merged to Onload-9 branch happy to update this: https://github.com/Xilinx-CNS/onload/commit/7764b681a8ccaa4385ac60375ca0a3f2dfb1313b

src/lib/zf/private/stack_alloc.c
ligallag-amd: I don't mind either option. If we always fail `ef_pio_alloc` when the PIO capability is not supported, I would rather we not make the additional call. But it doesn't matter that we do, it's just more work.

Slightly related, this error message only seems to refer to the `-ENOSYS` case - "the API does not know how to retrieve support for the supplied capability". I think this may be a tad misleading when reporting `-EOPNOTSUPP`. You have reported the rc which would dispel some confusion.

PR #69, opened 2025-02-21
ON-15637: Update zft_connect() doc to indicate the call does block

rdrinkwa-xilinx:
Previous documentation said the call didn't blocked but that was incorrect. Updated to state it will block when finding a route based on the value of 'arp_reply_timeout'.

### DIFF
--- a/src/include/zf/zf_tcp.h
+++ b/src/include/zf/zf_tcp.h
@@ -262,18 +262,23 @@ zft_addr_bind(struct zft_handle* handle, const struct sockaddr* laddr,
 ** If a specific local address has not been set via zft_addr_bind()
 ** then an appropriate one will be selected.
 **
-** This function does not block.  Functions that attempt to transfer data on
-** the zocket between zft_connect() and the successful establishment of the
-** underlying TCP connection will return an error.  Furthermore, failure of the
-** remote host to accept the connection will not be reported by this function,
-** but instead by any attempts to read from the zocket (or by zft_error()).  As
-** such, after calling zft_connect(), either
+** This function does not block to wait for a connection to be established.
+** Functions that attempt to transfer data on the zocket between zft_connect()
+** and the successful establishment of the underlying TCP connection will
+** return an error.  Furthermore, failure of the remote host to accept the
+** connection will not be reported by this function, but instead by any
+** attempts to read from the zocket (or by zft_error()).  As such, after
+** calling zft_connect(), either
 **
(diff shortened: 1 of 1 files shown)


src/include/zf/zf_tcp.h
```diff
@@ -262,19 +262,24 @@ zft_addr_bind(struct zft_handle* handle, const struct sockaddr* laddr,
 ** If a specific local address has not been set via zft_addr_bind()
 ** then an appropriate one will be selected.
 **
-** This function does not block.  Functions that attempt to transfer data on
-** the zocket between zft_connect() and the successful establishment of the
-** underlying TCP connection will return an error.  Furthermore, failure of the
-** remote host to accept the connection will not be reported by this function,
-** but instead by any attempts to read from the zocket (or by zft_error()).  As
-** such, after calling zft_connect(), either
+** This function does not block to wait for a connection to be established.
```
andresa-xilinx, line 265: This function does not block to wait for a connection to be established, but can block up to the ARP reply timeout.

andresa-xilinx: The bug title says "Update zft_connect doc to say it **does** block", but the way it is written although correct, indicates so too subtlety given that the potential blocking cause can be clearly pointed out.

rdrinkwa-xilinx: I can change the title to "the call may block" but I'm not sure what change you are suggesting I could make.

andresa-xilinx: I'd change the first sentence to "This function does not block to wait for a connection to be established, but can block up to the ARP reply timeout."

The change in bug title is also convenient.

pemberso-xilinx (approved):

jfeather-amd (approved):
Looks good, thanks! Based on your description, it sounds like this isn't something that changed recently - should this target v8_1 instead if the same applies there?

rdrinkwa-xilinx:
> Based on your description, it sounds like this isn't something that changed recently - should this target v8_1 instead if the same applies there?
The behaviour this is updated to reflect is present in v8_1 so I can change it so apply there instead if that's preferred.

jfeather-amd:
> The behaviour this is updated to reflect is present in v8_1 so I can change it so apply there instead if that's preferred.

Yes please, in general we merge forwards from the oldest to newest branches so it's best to target the oldest applicable branch for any change.

rdrinkwa-xilinx:
Rebase onto v8_1 done and confirmed only the intended changes.

PR #72, opened 2025-03-17
ON-13283: allow application to register callback for NIC reset

rdrinkwa-xilinx:
Add ability for an application to register a callback if the NIC used by the stack is reset. User is able to specify an argument to be passed to it. This allows the application to take action in that scenario.

Updated `efsink` to use the callback and show the interface name being passed through to the function.

### DIFF
--- a/src/lib/zf/private/reactor.c
+++ b/src/lib/zf/private/reactor.c
@@ -186,6 +186,13 @@ efct_pkt_memcpy(void* dst, const void* src, size_t n)
 }
 
 
+ZF_COLD static void
+zf_reactor_handle_reset_event(struct zf_stack* st)
+{
+  const struct zf_stack_impl* sti = ZF_CONTAINER(struct zf_stack_impl, st, st);
+  sti->reset_callback(sti->reset_callback_arg);
+}
+
 /* returns 1 if something interesting has happened */
 ZF_HOT int
 zf_reactor_process_event(struct zf_stack* st, int nic, ef_vi* vi, ef_event* ev)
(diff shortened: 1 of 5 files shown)


src/lib/zf/private/reactor.c
```diff
@@ -256,6 +263,9 @@ zf_reactor_process_event(struct zf_stack* st, int nic, ef_vi* vi, ef_event* ev)
   case EF_EVENT_TYPE_RX_REF_DISCARD:
     zf_reactor_handle_rx_ref_discard(st, nic, vi, ev);
     break;
+  case EF_EVENT_TYPE_RESET:
+    zf_reactor_handle_reset_event(st);
```
rhughes-xilinx, line 267: This is be called in an exciting context, i.e. with the stack mid-poll. If that's the way you want to do it then you'll have to write some really big disclaimers in the documentation about the things you are and are not permitted to do from within the callback.

rhughes-xilinx (commented):
Implementing it this way round seems dangerous. If we were to ship this feature as-is then customers might start writing code to reset zf stacks in the callback, which would then prevent us in the future from doing the stack reset internally

src/lib/zf/private/reactor.c
rdrinkwa-xilinx: The documentation could state that the user is not permitted to perform any actions on the stack, which I think would be recommended anyway.
Setting a flag within TCPDirect that triggers the callback after processing the list of events would be a change that could be implemented but I don't think it would really add that much safety to the stack.

sianj-xilinx: We already have a way of reporting events to be handled outside the context of a zf call using the zf_muxer. Have you looked at implementing this with a new event type that can be registered in the wait set and notified? Probably needs a bit of consideration to allow waiting on a stack rather than a socket, but the interface would be safer.

PR #75, opened 2025-04-03
ON-16435: Replace arch checks with nic capabilies

ligallag-amd:
```
ON-16435: Add getter for nic res flags

This makes it easier to change whether these flags are in zf_stack or
zf_stack_impl. Keeping them in the latter struct may have impact on
cache hit rate in the hotpath where these flags may be queried.
```

```
ON-16435: Add new ZF_RES_NIC_FLAGS and use them over arch checks
```

```
ON-16435: Add ef10ct arch check where capabilitiy is unclear
```

The final commit is probably not what we want but I am at a bit of a loss for what capabilities should be used in those cases. I therefore added ef10ct arch checks. Opening this as a draft because of this.

### Testing C-Model

`zfsink`
```
ZF_ATTR="interface=enp1s0f0" EF_VI_RXQ_SIZE=4096 EF_VI_PD_FLAGS=llct ./zfsink 192.168.135.105:2000
efct_design_parameters: unsupported rx_superbuf_bytes, 0 != 1048576
efct_design_parameters: rx_superbuf_bytes = 0 = 0, treating it as it were the standard value
# pkt-rate  bandwidth(Mbps)      total-pkts
        25                0               25
       398                0              423
---- Some time later (packet rate on C-Model is low) ----
       400                0             5120
       400                0             5520
       400                0             5920
        80                0             6000
         0                0             6000
```

`zfsend`

```
ZF_ATTR="interface=enp1s0f0" EF_VI_PD_FLAGS=llct ./zfsend -u -i 5000 192.168.135.105:2000 230.1.2.3:2000
```

All packets received at the other end. I had to modify zfsend to sleep 1000us between sends (not sure of the sends are batched) because otherwise the sends would fail and zfsend would hit an assertion. This is fine because the C-Model cannot keep up with the sends.

### DIFF
--- a/src/include/zf_internal/tx_send.h
+++ b/src/include/zf_internal/tx_send.h
@@ -288,6 +288,7 @@ zf_send(struct zf_tx* restrict tx,
     }
   }
   zf_assert_nequal(vi->nic_type.arch, EF_VI_ARCH_EFCT);
+  zf_assert_nequal(vi->nic_type.arch, EF_VI_ARCH_EF10CT);
   /* Strictly speaking, we should call ef_vi_transmit_ctpio_fallback()
    * to post a CTPIO fallback descriptor. However, that entry point is
    * actually identical to this one, and this code path is also used
--- a/src/include/zf_internal/tx_send.h
+++ b/src/include/zf_internal/tx_send.h
@@ -164,7 +164,7 @@ zf_send(struct zf_tx* restrict tx,
     return -EAGAIN;
 #endif
 
-  if( vi->nic_type.arch == EF_VI_ARCH_EFCT ) {
+  if( *zf_stack_res_nic_flags(st, nicno) & ZF_RES_NIC_FLAG_CTPIO_ONLY ) {
     /* FIXME get X3 overhead sorted properly */
     if(ZF_UNLIKELY( ef_vi_transmit_space_bytes(vi) < (int)tot_len + 128 )) {
       if( st->flags & ZF_STACK_FLAG_TRANSMIT_WARM_ENABLED )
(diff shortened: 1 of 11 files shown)


src/lib/zf/private/reactor.c
```diff
@@ -508,10 +508,11 @@ __zf_reactor_perform(struct zf_stack* st, unsigned spin_cnt)
            * from the future in the next RX buffer. */
           pkt_id next_packet_id = 0;
           char* next_packet;
-          bool is_efct = vi->nic_type.arch == EF_VI_ARCH_EFCT;
+          bool is_rx_ref = *zf_stack_res_nic_flags(st, nic) &
```
sianj-xilinx, line 511: You're using these flags on the critical path. The `zf_stack_impl` is used for cold path state. If we need the checks on the hot path that information should move to `zf_stack_nic`. That might just be the RX_REF capability - probably needs a bit of care to ensure that this is done in a way that doesn't impact performance.

src/include/zf_internal/tx_send.h
```diff
@@ -288,6 +288,7 @@ zf_send(struct zf_tx* restrict tx,
     }
   }
   zf_assert_nequal(vi->nic_type.arch, EF_VI_ARCH_EFCT);
+  zf_assert_nequal(vi->nic_type.arch, EF_VI_ARCH_EF10CT);
```
sianj-xilinx, line 291: Assert that CTPIO_ONLY capability wasn't set?

src/lib/zf/tx_warm.c
```diff
@@ -16,7 +16,8 @@ int enable_tx_warm(struct zf_tx* tx, zf_tx_warm_state* state)
   zf_log_stack_trace(st, "%s: TX warm enabled\n", __func__);
   char* ctpio_warm_buf = NULL;
   state->ctpio_warm_buf_id = PKT_INVALID;
-  if( vi->vi_flags & EF_VI_TX_CTPIO && vi->nic_type.arch != EF_VI_ARCH_EFCT ) {
+  if( vi->vi_flags & EF_VI_TX_CTPIO && vi->nic_type.arch != EF_VI_ARCH_EFCT
+      && vi->nic_type.arch != EF_VI_ARCH_EF10CT ) {
```
sianj-xilinx, line 20: The check here is on whether we need to post a dma fallback, which we don't if we have the CTPIO_ONLY capability, so I think that's what we should use.

PR #84, opened 2025-07-01
ON-16670: Future-proof LTO

ivatet-amd:
This patch does not enable LTO, but facilitates it later.

(Happy to drop the PR if we think it is confusing!)


ivatet-amd:
On second thought, changes covered in these patches might be LTO implementation-specific and become irrelevant in the newer versions of GCC. Closing.

PR #100, opened 2026-01-13
ON-17116: disable overlapped receive on unsupported EFCT architectures

mseymour-xilinx:
Overlapped receive would fail badly on unsupported architectures, where we don't have access to the packet buffers to poison them appropriately before posting. Without the poison we would report that all data was available immediately, when it might still be in transit.

Don't report ZF_EPOLLIN_OVERLAPPED in this case. We will still handle future packets internally, but the app will receive the data normally when the packet has completed.

### DIFF
--- a/src/lib/zf/private/reactor.c
+++ b/src/lib/zf/private/reactor.c
@@ -643,6 +643,7 @@ __zf_reactor_perform(struct zf_stack* st, unsigned spin_cnt)
              * but it's advantageous to set these sooner rather than later. */
             st->future_nic_id = nic;
             st->future_packet_id = next_packet_id;
+            st->pftf.can_overlap = ! is_efct;
 
             /* A packet is arriving from the future.  We can begin protocol-
              * processing in the hope that the rest of the packet arrives
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,7 +1,7 @@
 tcpdirect-9.1.1
 ---------------
 
-No changes since previous version
+ON-17116: disable overlapped receive on unsupported EFCT architectures
 
 
 tcpdirect-9.1.0.4
--- a/src/include/zf_internal/private/zf_stack_def.h
+++ b/src/include/zf_internal/private/zf_stack_def.h
@@ -120,6 +120,7 @@ struct zf_stack {
      * Typically set when unrelated user visible event interrupted pftf.
      */
     unsigned event_occurred_carry;
+    bool can_overlap;
   } pftf;
 
   bool x4_shared_mode;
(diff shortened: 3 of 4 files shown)


mseymour-xilinx:
Testing: nobbled zfsend to send a known data pattern; zfsink to check the pattern and to attempt and report overlapped receives.

Before (failing):
```
$ zfsink -m dellr650g-l:12345
# pkt-rate  bandwidth(Mbps)       total-pkts      overlapped
         0                0                0                0
         0                0                0                0
         0                0                0                0
*** TEST SEQ               ff: 00 != 1f
```

After (succeeding without overlap):
```
$ zfsink -m dellr650g-l:12345
# pkt-rate  bandwidth(Mbps)       total-pkts      overlapped
         0                0                0                0
         0                0                0                0
         0                0                0                0
         0                0                0                0
    123545              506           123545                0
    163253              668           286798                0
    162646              666           449444                0
    162767              666           612211                0
    162667              666           774878                0
    163007              667           937885                0
     62115              254          1000000                0
         0                0          1000000                0
         0                0          1000000                0
         0                0          1000000                0
```
Enterprise still works with overlap:
```
$ ZF_ATTR="$ZF_ATTR;rx_datapath=enterprise;pio=0" zfsink -m dellr650g-l:12345
# pkt-rate  bandwidth(Mbps)       total-pkts      overlapped
         0                0                0                0
         0                0                0                0
         0                0                0                0
         0                0                0                0
     67758              277            67758            67235
    163819              671           231577           229885
    163483              669           395060           392200
    162996              667           558056           554003
    162987              667           721043           715806
    162908              667           883951           877469
    116049              475          1000000           992641
         0                0          1000000           992641
         0                0          1000000           992641
```

src/lib/zf/private/reactor.c
```diff
@@ -643,6 +643,7 @@ __zf_reactor_perform(struct zf_stack* st, unsigned spin_cnt)
              * but it's advantageous to set these sooner rather than later. */
             st->future_nic_id = nic;
             st->future_packet_id = next_packet_id;
+            st->pftf.can_overlap = ! is_efct;
```
jfeather-amd, line 646: It would be nice if we could get this information from somewhere else, like the capabilities API, instead of making an assumption about EFCT here, but I think there are already a few places that do similar things so don't think this should block this change.

jfeather-amd (approved):
Looks good to me

src/lib/zf/private/reactor.c
mseymour-xilinx: I agree, but I think that's out of scope for this bug fix, especially as there's already a dodgy architecture check in this function.

sianj-xilinx: That was my thought process too!

sianj-xilinx (approved):

PR #101, opened 2026-01-28
ON-17181: updated to copy packages to artifactory

haria-xilinx:
Updated test.groovy and pipeline.groovy to build and upload packages to artifacts at "tcpdirect/solarflare"
tested it. it looks fine
https://xcb-jenkins-onload.xilinx.com/blue/organizations/jenkins/tcpdirect%2Ftcpdirect_pipeline/detail/v9_1/33/pipeline

### DIFF
--- a/ci/pipeline.groovy
+++ b/ci/pipeline.groovy
@@ -39,17 +39,19 @@ class ZfPipeline implements Serializable {
     script.archiveArtifacts(zipfile)
   }
 
-  void doPipeline(String tcpdirect_stash,
-                  String onload_stash,
-                  String onload_tarball_stash,
-                  String onload_tarball,
-                  version_info)
+  Map doPipeline(String tcpdirect_stash,
+                 String onload_stash,
+                 String onload_tarball_stash,
+                 String onload_tarball,
+                 version_info)
   {
+    def package_locations = [:]
+
     script.stage('Parallel builds') {
       utils.parallel(
         'build ndebug': {
           script.node('unit-test-master') {
-            script.sh 'rm -fr tcpdirect onload'
+            script.deleteDir()
             script.unstash(tcpdirect_stash)
             script.unstash(onload_stash)
             script.sh 'ls -lad $PWD'
@@ -68,7 +70,7 @@ class ZfPipeline implements Serializable {
         },
         'build libzf_sockets': {
           script.node('unit-test-master') {
-            script.sh 'rm -fr tcpdirect onload'
+            script.deleteDir()
             script.unstash(tcpdirect_stash)
             script.unstash(onload_stash)
             script.sh 'ls -lad $PWD'
@@ -85,7 +87,7 @@ class ZfPipeline implements Serializable {
         },
         'run tests': {
           script.node('unit-test-master') {
-            script.sh 'rm -fr tcpdirect onload packetdrill-tcpdirect test-results'
+            script.deleteDir()
             script.unstash(tcpdirect_stash)
             script.unstash(onload_stash)
             script.sh 'ls -lad $PWD'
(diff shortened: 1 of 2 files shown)


haria-xilinx:
Sorry missed out some of commneted code for testing. now updated

ci/pipeline.groovy
```diff
@@ -111,39 +113,50 @@ class ZfPipeline implements Serializable {
...
-        }
-      }
-    }
-
     script.stage('Build packages') {
       utils.parallel(
-        "publish release rpm": {
+        "build tarball": {
+          script.node('unit-test-master') {
+            script.sh 'rm -fr tcpdirect onload'
```
jfeather-amd: It looks like this directory is accumulating `onload-<version>.tgz` files (unsure if this was the case before this change). Possibly we should copy what other steps in the pipeline do and replace this manual `rm` with `script.deleteDir()`.

ci/pipeline.groovy
```diff
@@ -111,39 +113,50 @@ class ZfPipeline implements Serializable {
...
+            script.dir('tcpdirect/build/') {
+              def tarball = "tcpdirect-${version_info.TCPDIRECT_VERSION}.tgz"
+              def tarball_md5 = "${tarball}.md5"
+
+              script.archiveArtifacts(artifacts: '*.tgz')
+              script.archiveArtifacts(artifacts: '*.md5')
+              script.archiveArtifacts(artifacts: '*.txt')
+
+              package_locations['raw_tarball'] = tarball
+              package_locations['tarball_md5'] = tarball_md5
```
jfeather-amd: nit: I think it might be nicer to keep the prefix identical for the md5, so have `raw_tarball` and `raw_tarball_md5` or `tarball` and `tarball_md5` instead of a mix between the two

ci/pipeline.groovy
```diff
@@ -111,39 +113,50 @@ class ZfPipeline implements Serializable {
...
+
+              script.sh 'rm *ReleaseNotes.txt'
+              script.stash name: 'text_files', includes: '*.txt'
+              zip_and_archive_files(
+                "tcpdirect-${version_info.TCPDIRECT_VERSION}-tarball-doxbox.zip",
+                '*.tgz',
+                '*.md5',
+                '*.txt'
+              )
+              def tarball_doxbox = "tcpdirect-${version_info.TCPDIRECT_VERSION}-tarball-doxbox.zip"
```
jfeather-amd: Probably worth reusing the tarball file name in the creation as well, e.g.:
```java
def tarball_doxbox = "tcpdirect-${version_info.TCPDIRECT_VERSION}-tarball-doxbox.zip"
zip_and_archive_files(
  tarball_doxbox,
  '*.tgz',
  '*.md5',
  '*.txt'
)
```

ci/pipeline.groovy
```diff
@@ -153,42 +166,77 @@ class ZfPipeline implements Serializable {
...
+            script.stash name: 'tcpdirect-tar', includes: "tcpdirect-${version_info.TCPDIRECT_VERSION}.tar.gz"
+            script.sh "mv tcpdirect-${version_info.TCPDIRECT_VERSION}.tar.gz tcpdirect-source-${version_info.TCPDIRECT_VERSION}.tar.gz"
+            def source = "tcpdirect-source-${version_info.TCPDIRECT_VERSION}.tar.gz"
+            package_locations['source'] = source
+            script.stash name: 'source', includes: source
+            script.archiveArtifacts allowEmptyArchive: true, artifacts: source, followSymlinks: false
+
+            script.deleteDir()
+            script.unstash('tcpdirect-tar')
+            script.unstash(tcpdirect_stash)
```
jfeather-amd: I see you've merged what were previously two sections; we can probably avoid stashing, unstashing, and destroying the workspace in between now. It would need the above `mv` to become a `cp` (or later references to that file to be updated).

ci/pipeline.groovy
```diff
@@ -153,42 +166,77 @@ class ZfPipeline implements Serializable {
...
+          }
+        },
+        "build deb": {
+          script.node("deb") {
+            script.deleteDir()
+            script.unstash(tcpdirect_stash)
+            script.sh "tcpdirect/scripts/zf_mksrc --version ${version_info.TCPDIRECT_VERSION}"
+            script.stash name: 'tcpdirect-tar', includes: "tcpdirect-${version_info.TCPDIRECT_VERSION}.tar.gz"
+            script.sh "mv tcpdirect-${version_info.TCPDIRECT_VERSION}.tar.gz tcpdirect-source-${version_info.TCPDIRECT_VERSION}.tar.gz"
+            def source = "tcpdirect-source-${version_info.TCPDIRECT_VERSION}.tar.gz"
```
jfeather-amd: Swapping these and reusing `source` in the `mv` would be desirable

ci/pipeline.groovy
```diff
@@ -153,42 +166,77 @@ class ZfPipeline implements Serializable {
...
+            def safe_version_string = version_info.TCPDIRECT_VERSION.replace('-','_')
+            def srpm_file = "rpmbuild/SRPMS/tcpdirect-${safe_version_string}-1.src.rpm"
             zip_and_archive_files(
               "tcpdirect-${version_info.TCPDIRECT_VERSION}-srpm-doxbox.zip",
-              'rpmbuild/SRPMS/*.src.rpm',
+              srpm_file,
               '*.txt'
             )
-            script.sh "rm -rf ${outdir}"
+            def srpm_doxbox = "tcpdirect-${version_info.TCPDIRECT_VERSION}-srpm-doxbox.zip"
```
jfeather-amd: It'd also be good to use `srpm_doxbox` in its creation above.

ci/test.groovy
```diff
@@ -75,6 +76,28 @@ nm.slack_notify() {
   am.doAutosmoke(version_info.ONLOAD_REPO_SOURCE, version_info.ONLOAD_VERSION,
                  'last_known_good/' + version_info.ONLOAD_VERSION,
                  orgfilesRepo, personal_job, true, am.sourceUrl(), env.BRANCH_NAME)
+
+  // Publish packages to Artifactory
+  echo "Echo Branch: ${env.BRANCH_NAME}, Echo Personal job: ${personal_job}"
+  def VERSION = version_info.TCPDIRECT_VERSION
+  def short_revision = scmVars.GIT_COMMIT.substring(0,12)
+  def long_revision = scmVars.GIT_COMMIT
+  echo "VERSION: ${VERSION}, revision: ${long_revision}"
```
jfeather-amd: This is all outside of a stage, so won't play nicely with the blue ocean view.

ci/test.groovy
```diff
@@ -75,6 +76,28 @@ nm.slack_notify() {
...
+
+  node('master') {
+    stage('Publish to Artifactory') {
+      utils.withArtifactoryURL() {
+        utils.withArtifactoryCreds() {
+          def publisher = new ArtifactoryPublisher(this)
```
jfeather-amd: It seems to be more standard within our groovy files that we declare these helpers at the top.

ci/test.groovy
```diff
@@ -75,6 +76,28 @@ nm.slack_notify() {
...
+          echo "Publishing TCPDirect packages to Artifactory..."
+          def branch_name = env.BRANCH_NAME.toLowerCase().replaceAll('^v', '')
+          def branch_with_product = "tcpdirect-${branch_name}"
```
jfeather-amd: I think we probably want to be using the `version_info.TCPDIRECT_VERSION` here, as that'll give us the full version number from the `versions.env` file.

jfeather-amd (commented):
I've left a few comments, but if you could, it'd make this easier to review if you split up the changes into multiple commits. For example, you move code around in a few places without changing it, which probably should be its own commit.

ci/pipeline.groovy
haria-xilinx: It is not introduced by me. any how thank you josh for suggesting cleaner method.

ci/pipeline.groovy
haria-xilinx: even noo need to copy it. we can use source directly.

ci/test.groovy
haria-xilinx: I have followed onload version representation without 'v'. So Runbench can understand it while down loading.

haria-xilinx:
jenkins run after updates
https://xcb-jenkins-onload.xilinx.com/blue/organizations/jenkins/tcpdirect%2Ftcpdirect_pipeline/detail/v9_1/39/pipeline

jfeather-amd (approved):
Thanks Hari, looks good. Before merging I think this will need rebasing to target the `v8_1` branch. I would also like to see what happens with the publishing step on the master branch, so it would be good if you could give that a go

haria-xilinx:
This is already in sync with master. I will merge with master then we can rabse it with other branches. is it OK?

haria-xilinx:
On master also works fine
https://xcb-jenkins-onload.xilinx.com/blue/organizations/jenkins/tcpdirect%2Ftcpdirect_pipeline/detail/master/449/pipeline

jfeather-amd:
> This is already in sync with master. I will merge with master then we can rabse it with other branches.

It would be more typical to base a change on the oldest applicable version, then merge the change forwards to the other branches. It's possible to merge to master and cherry-pick the change to the older branches, but I don't think that's the preferred approach.

> On master also works fine

Thanks for testing. Is the result as you expect in artifactory? For example, the path to the SRPM appears to be "tcpdirect-master/tcpdirect-srpm/9.2.0.449/tcpdirect-srpm-9.2.0.449.rpm", would we want to mix "master" and "9.2.0" versioning?

haria-xilinx:
> > This is already in sync with master. I will merge with master then we can rabse it with other branches.
> 
> It would be more typical to base a change on the oldest applicable version, then merge the change forwards to the other branches. It's possible to merge to master and cherry-pick the change to the older branches, but I don't think that's the preferred approach.
> 
> > On master also works fine
> 
> Thanks for testing. Is the result as you expect in artifactory? For example, the path to the SRPM appears to be "tcpdirect-master/tcpdirect-srpm/9.2.0.449/tcpdirect-srpm-9.2.0.449.rpm", would we want to mix "master" and "9.2.0" versioning?

No each branch separately stores
tcpdirect-master->tcpdirect-doxbox-srpm>9.2.0.449->tcpdirect-srpm-9.2.0.449.rpm
tcpdirect-9.1->tcpdirect-doxbox-srpm>9.2.0.40->tcpdirect-srpm-9.2.0.40.rpm

PR #106, opened 2026-03-31
ON-17171: set the max number of evq events in zf_emu

jfeather-amd:
This is required to match a change in onload that adds accounting of the number of EVQ slots that remain unfilled to ensure we avoid an overflow in our EVQ.

### Testing Done
Unit tests pass when compiled again the pending onload changes.


mseymour-xilinx (approved):

jfeather-amd:
Force push rebases to v9_2.

